PR #11, opened 2015-11-03
updated prerequisite and build steps in readme.md

danielli90:


danielli90:
need to squash all 7 commits into one

PR #14, opened 2015-11-04
capturing stacktrace when exception occur in JVM & setting app name

skaarthik:


skaarthik:
will send another PR from my master repo

PR #16, opened 2015-11-05
Fix minor issues in code-docs

jthelin:
Fix some minor issues in code-docs and messages.

- Fix various minor issues with XML code-docs.

- Mostly required replacing < with &lt; for valid XML.

- Fix minor typo in comment.

- Fix wrong parameters being passed into string.Format

### DIFF
--- a/csharp/Adapter/Microsoft.Spark.CSharp/Sql/DataFrame.cs
+++ b/csharp/Adapter/Microsoft.Spark.CSharp/Sql/DataFrame.cs
@@ -221,22 +221,22 @@ namespace Microsoft.Spark.CSharp.Sql
         }
 
         /// <summary>
-        /// Join with another DataFrame
+        /// Join with another DataFrame - Cartesian join
         /// </summary>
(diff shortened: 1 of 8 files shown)


csharp/Adapter/Microsoft.Spark.CSharp/Sql/DataFrame.cs
```diff
@@ -149,7 +148,7 @@ public DataFrame Select(string firstColumnName, params string[] otherColumnNames
         ///   df.SelectExpr("colA", "colB as newName", "abs(colC)")
         ///   
         /// </summary>
-        /// <param name="columnNames"></param>
+        /// <param name="columns"></param>
```
skaarthik: columnExpressions?

jthelin: Good catch @skaarthik - param name was also changed in commit 037d183acb9490f929f7971df5b6754b3908a352. Fixed and repushing.

PR #32, opened 2015-11-11
Minor improvements to handling some error conditions.

jthelin:
Minor improvements to handling some error conditions and first-use confusion points.

Improved error reporting of missing `CSharpWorkerPath` App.config value.

- Throw `ConfigurationErrorsException` with more specific message about missing `CSharpWorkerPath` value in `App.config`.

~~For Local configuration, use the file name part of the `CSharpWorkerPath` value, if specified,
  otherwise use the default value `CSharpWorker.exe` from base class.~~

Guard against `NullReferenceException` in `Dispose()`.

~~Add `GetJavaPort` method to prompt user for port number and read console.~~

### DIFF
--- a/csharp/Adapter/Microsoft.Spark.CSharp/Configuration/ConfigurationService.cs
+++ b/csharp/Adapter/Microsoft.Spark.CSharp/Configuration/ConfigurationService.cs
@@ -153,25 +153,37 @@ namespace Microsoft.Spark.CSharp.Configuration
                 : base(configuration)
             {}
 
-            internal override string GetCSharpWorkerPath()
+            private string workerPath;
+            internal override string GetCSharpRDDExternalProcessName()
             {
+                // SparkCLR jar and driver, worker & dependencies are shipped using Spark file server. 
+                // These files are available in the Spark executing directory at executor node.
+
+                if (workerPath != null) return workerPath; // Return cached value
+
                 KeyValueConfigurationElement workerPathConfig = appSettings.Settings["CSharpWorkerPath"];
-                string workerPath;
                 if (workerPathConfig == null)
                 {
                     // Path for the CSharpWorker.exe was not specified in App.config
                     // Try to work out where location relative to this class.
                     // Construct path based on well-known file name + directory this class was loaded from.
-                    string procFileName = GetCSharpRDDExternalProcessName();
+                    string procFileName = base.GetCSharpRDDExternalProcessName();
                     string procDir = Path.GetDirectoryName(GetType().Assembly.Location);
                     workerPath = Path.Combine(procDir, procFileName);
-                    //throw new ConfigurationErrorsException("Need to set CSharpWorkerPath value in App.config");
+                    logger.LogDebug("Using synthesized value for CSharpWorkerPath : " + workerPath);
                 }
                 else
                 {
                     // Explicit path for the CSharpWorker.exe was listed in App.config
                     workerPath = workerPathConfig.Value;
+                    logger.LogDebug("Using CSharpWorkerPath value from App.config : " + workerPath);
                 }
+                return workerPath;
+            }
+
+            internal override string GetCSharpWorkerPath()
+            {
+                string workerPath = GetCSharpRDDExternalProcessName();
                 return new Uri(workerPath).ToString();
             }
         }
--- a/csharp/Adapter/Microsoft.Spark.CSharp/Configuration/ConfigurationService.cs
+++ b/csharp/Adapter/Microsoft.Spark.CSharp/Configuration/ConfigurationService.cs
@@ -153,16 +153,15 @@ namespace Microsoft.Spark.CSharp.Configuration
                 : base(configuration)
             {}
 
-            internal override string GetCSharpRDDExternalProcessName()
-            {
-                return appSettings.Settings["CSharpWorkerPath"].Value;
-            }
-            
             internal override string GetCSharpWorkerPath()
             {
-                return new Uri(appSettings.Settings["CSharpWorkerPath"].Value).ToString();
+                KeyValueConfigurationElement workerPathConfig = appSettings.Settings["CSharpWorkerPath"];
+                if (workerPathConfig == null)
+                {
+                    throw new ConfigurationErrorsException("Need to set CSharpWorkerPath value in App.config");
+                }
+                return new Uri(workerPathConfig.Value).ToString();
             }
-
         }
 
         /// <summary>
@@ -177,7 +176,12 @@ namespace Microsoft.Spark.CSharp.Configuration
 
             internal override int GetPortNumber()
             {
-                var cSharpBackendPortNumber = int.Parse(appSettings.Settings["CSharpBackendPortNumber"].Value);
+                KeyValueConfigurationElement portConfig = appSettings.Settings["CSharpBackendPortNumber"];
+                if (portConfig == null)
+                {
+                    throw new ConfigurationErrorsException("Need to set CSharpBackendPortNumber value in App.config");
+                }
+                int cSharpBackendPortNumber = int.Parse(portConfig.Value);
                 logger.LogInfo(string.Format("CSharpBackend port number read from app config {0}", cSharpBackendPortNumber));
                 return cSharpBackendPortNumber;
             }
@@ -196,7 +200,6 @@ namespace Microsoft.Spark.CSharp.Configuration
             {
                 return "CSharpSparkWorker.exe";
             }
-
         }
     }
 
--- a/csharp/Adapter/Microsoft.Spark.CSharp/Configuration/ConfigurationService.cs
+++ b/csharp/Adapter/Microsoft.Spark.CSharp/Configuration/ConfigurationService.cs
@@ -156,11 +156,23 @@ namespace Microsoft.Spark.CSharp.Configuration
             internal override string GetCSharpWorkerPath()
             {
                 KeyValueConfigurationElement workerPathConfig = appSettings.Settings["CSharpWorkerPath"];
+                string workerPath;
                 if (workerPathConfig == null)
                 {
-                    throw new ConfigurationErrorsException("Need to set CSharpWorkerPath value in App.config");
+                    // Path for the CSharpWorker.exe was not specified in App.config
+                    // Try to work out where location relative to this class.
+                    // Construct path based on well-known file name + directory this class was loaded from.
+                    string procFileName = GetCSharpRDDExternalProcessName();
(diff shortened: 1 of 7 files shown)


csharp/Adapter/Microsoft.Spark.CSharp/Configuration/ConfigurationService.cs
```diff
@@ -150,14 +150,25 @@ internal SparkCLRLocalConfiguration(System.Configuration.Configuration configura
 
             internal override string GetCSharpRDDExternalProcessName()
             {
-                return appSettings.Settings["CSharpWorkerPath"].Value;
+                KeyValueConfigurationElement workerPathConfig = appSettings.Settings["CSharpWorkerPath"];
+                if (workerPathConfig != null)
+                {
+                    // Use the file name part of the CSharpWorkerPath value, if specified
+                    return Path.GetFileName(workerPathConfig.Value);
```
jthelin: This is the only real functional change, but seems like this is the intended semantics for the `GetCSharpRDDExternalProcessName` method?

csharp/Worker/Microsoft.Spark.CSharp/Worker.cs
```diff
@@ -36,8 +36,9 @@ public class Worker
         static void Main(string[] args)
         {
             PrintFiles();
-            int javaPort = int.Parse(Console.ReadLine());
```
skaarthik: This is not an interactive scenario. Port number is written to output stream by https://github.com/apache/spark/blob/master/core/src/main/scala/org/apache/spark/api/python/PythonWorkerFactory.scala#L125 and read here.

csharp/Adapter/Microsoft.Spark.CSharp/Configuration/ConfigurationService.cs
skaarthik: For SparkCLRLocalConfiguration, CSharpWorkerPath is a required parameter in appSettings. If it is not set, we should just throw exception. Using base.GetCSharpRDDExternalProcessName value would lead to exception later.

csharp/Worker/Microsoft.Spark.CSharp/Worker.cs
jthelin: OK. Thanks for the pointer to the "other side". :)
I'll revert this part.

csharp/Adapter/Microsoft.Spark.CSharp/Configuration/ConfigurationService.cs
jthelin: My question is prompted by what is the right value to be returned from GetCSharpRDDExternalProcessName.

Shouldn't *ProcessPath return the full path to the .exe and *ProcessName just the "file" part?
For example: `C:\my\path\CSharpWorker.exe` and `CSharpWorker.exe` respectively?

If that is the case, then can't *ProcessName automatically be derived using Path.GetFileName on the full path value returned from GetCSharpWorkerPath reading App.config setting?

I don't think I have quite grokked the difference between the expected config for "local" and "debug" execution modes.

jthelin: I guess the other way of asking this is: semantically won't `GetCSharpRDDExternalProcessName` always have the value `CSharpWorker.exe` or can it really be changed by App.config?

jthelin:
Addressed one review comment, and rebased against master 1d3eb86137eb712405ba955b9dd4373adfcb38a3

csharp/Adapter/Microsoft.Spark.CSharp/Configuration/ConfigurationService.cs
jthelin: Actually, the more I look at this, seems that GetCSharpWorkerPath method is never used, so maybe should just rename GetCSharpRDDExternalProcessName?

danielli90:
LTGM. Thanks @jthelin

skaarthik:
@jthelin I will take a look at the recent iteration. Have you verified your changes by running all the samples?

xiongrenyi:
also looks good to me

jthelin:
I am still having problems getting the samples to run this morning, with same issue as reported in #35 
```
Error: Could not find or load main class org.apache.spark.launcher.SparkCLRSubmitArguments
```

Please hold off merging this PR until I can confirm it runs smoothly against master HEAD, as @skaarthik  suggested.

jthelin:
OK, i think it is safe to merge this change now.

Most of the samples run ok on my machine, which suggests that the startup code this change affects is working as intended.

Some of the samples do fail when i run them, but with weird internal errors on callbacks between CSharpWorker and JVM, which i am pretty sure are unrelated to this change and need to be investigated separately.
```
collectAndServe on object of type NullObject failed
```

HTH.

skaarthik:
@jthelin When you run samples on a repo without your changes,do you see the same behavior? @danielli90 or @xiongrenyi can you help @jthelin successfully run the samples? If samples are failing, I would like them fixed before merging any new PRs.

jthelin:
Hmm. Your hunch is right @skaarthik -- there is something quite strange happening with this PR.

I worked out a few of the issues I hit, but not totally got to a clean run yet.

1. My `CSharpWorkerPath` entry in `SparkCLRSamples.exe.config` was pointing to the copy of `CSharpWorker.exe` in the `bin` directory and not the `samples` directory. Changing this got rid of the of the NullObject error I mentioned above.
2. I did not have copies of the extra csv libraries required for the DataFrame TextFile API.

In an attempt to create a repeatable runtime environment, I created a meta-build script (submitted as PR #45) to assemble the runtime artifacts in a consistent manner. 

@danielli90 / @xiongrenyi - Could you please take a look at PR #45 to see if my runtime environment looks correct.

I also made a few tweaks to the Samples driver program (see PR #46) to make it easier to track outcome of each sample, and report final pass/fail numbers.

danielli90:
Reviewing #45 and #46 now.

danielli90:
#45 and #46 are merged. @jthelin, can you confirm you are not running issues in running samples locally with this PR? If yes, we can go ahead merge this PR; If not, you can ask @tawan0109 to buddy-test this PR for you.

jthelin:
Yes, unfortunately I am still having problems getting a clean run of Samples for this PR. :(

@tawan0109 - could you take a look at what is happening deep inside the JVM Bridge code when this PR is applied? Thanks -- extra eyes always appreciated! :)

jthelin:
Rebased against master HEAD 7d62d38ee6add997a77bef3fa9ea6b63f3afb179

tawan0109:
@jthelin checking

tawan0109:
@jthelin when constructing a org.apache.spark.api.csharp.CSharpRDD object in JVM, the 6th parameter is required to be the cSharpWorkerExecutable name, in the changes made to 
```
csharp/Adapter/Microsoft.Spark.CSharp/Configuration/IConfigurationService.cs 
```
and

```
csharp/Adapter/Microsoft.Spark.CSharp/Proxy/Ipc/SparkContextIpcProxy.cs, 
```
CSharpWorkerPath is given instead of the CSharpWorker name. This is might be the root cause of the samples failure.

```
                 new object[]  
              {  
                     rdd, command, hashTableReference, arrayListReference, preservesPartitioning,   
 -                   SparkCLREnvironment.ConfigurationService.GetCSharpRDDExternalProcessName(),  
 +                   SparkCLREnvironment.ConfigurationService.GetCSharpWorkerPath(),  
                     "1.0",  
                     jbroadcastVariables, jvmAccumulatorReference  
                 });  

```

jthelin:
To clarify my understanding: are you saying that value passed for that sixth parameter should be just the "filename" part of the path (eg: "CSharpWorker.exe") rather than the full path (eg: "C:\my\path\CSharpWorker.exe")

From code search on the C# side, it seemed that only one of those methods (`GetCSharpRDDExternalProcessName` / `GetCSharpWorkerPath`) was ever used. 
I guess I picked the wrong way to jump on that topic :(

Thanks for investigating the issue @tawan0109 -- much appreciated.

tawan0109:
For local mode, GetCSharpRDDExternalProcessName() and GetCSharpWorkerPath() returns different things: 

GetCSharpRDDExternalProcessName() returns raw string from App.config
```
D:/SparkCLR/Release/samples/CSharpWorker.exe
```
GetCSharpWorkerPath: Uri.toString()
```
file:///D:/SparkCLR/Release/samples/CSharpWorker.exe
```

jthelin:
But the confusing part is that (as far as I can tell) `GetCSharpWorkerPath` method is never called from anywhere?

Also, the skeleton code in `SparkCLRYarnClusterConfiguration` for `GetCSharpWorkerPath` method returns the "raw" exe-name string (no path and not Uri'ed) so their is some inconsistency there.

I am pretty sure it would be better to have a strongly-typed method returning Uri for when/where a Uri'ed value is required, and otherwise just use full-path(string) in the other places.

Is there anywhere on the Scala side that only uses short (file) name rather than full paths?
Or somewhere that requires the Uri form of the path?
I can't seem to find any, although much of the Scala side of the code has documentation measured on a negative scale! :(

BTW, I think I have managed to get things unraveled now so that the Samples can run in local mode with zero config -- the location of the CSharpWorker.exe can be inferred from the location of the Configuration class(es). Saves a lot of hassle! :)

I want to sleep on this and take another look at cleaning up the commit in the morning before any final merge, so consider the updates I just pushed as slightly rough draft and advisory.

jthelin:
The Samples all run ok for me with db2c445a20860af07b885a20843bd3a5258becfe except that:

- `DStreamTextFileSamples` did not finish shutdown, and blocked the rest of the run. I have to comment out the [Sample] attribute on that method for the rest of the Samples suite to run.

- `DFTextFileJoinTempTableSample` and `DFTextFileJoinTableDSLSample` both fail with "loadTextFile on object of type NullObject failed" error from the Scala side. They are both reported in the end-of-run completion report as failing, but don't block the suite.

All three of these problems occur on my machine both with and without this PR, so I believe they are separate issues from these changes.

tawan0109:
- `DStreamTextFileSamples` This is a known issue, Actually if you run this sample standalone without others, it will succeed.
- ` DFTextFileJoinTempTableSample` and ` DFTextFileJoinTableDSLSample `, please make sure environment variable `SPARKCSV_JARS` is set properly.

danielli90:
I patched this #32 on top of the latest codebase - it passed all samples locally. There are exceptions at the end of sample data run in deleting temp directories - but that is common before applying #32 patch. I will merge this #32 in.

PR #44, opened 2015-11-12
improve DStreamSample's synchronization of C# and Java to let StreamingContext stop gracefully

xiongrenyi:
it's a minor change in a streaming sample, would like code review on PR directly

### DIFF
--- a/csharp/Samples/Microsoft.Spark.CSharp/DStreamSamples.cs
+++ b/csharp/Samples/Microsoft.Spark.CSharp/DStreamSamples.cs
@@ -40,6 +40,7 @@ namespace Microsoft.Spark.CSharp
                     }
                     System.Threading.Thread.Sleep(200);
                 }
+
                 foreach (var file in Directory.GetFiles(testDir, "*"))
                     File.Delete(file);
             });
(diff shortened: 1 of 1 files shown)


csharp/Samples/Microsoft.Spark.CSharp/DStreamSamples.cs
```diff
@@ -83,6 +88,9 @@ internal static void DStreamTextFileSamples()
                 System.Threading.Thread.Sleep(1000);
             }
 
+            // wait ForeachRDD to complete to let ssc.Stop() gracefully
+            System.Threading.Thread.Sleep(2000);
```
skaarthik: Are we recommending to follow the same practice when they write their streaming apps in SparkCLR?

danielli90:
LGTM.

It will be good eventually to have a DStream sample similar to ones in Scala, so the readers can understand that SparkCLR doesn't introduce any new programming model.

csharp/Samples/Microsoft.Spark.CSharp/DStreamSamples.cs
xiongrenyi: actually not, streaming has a separate callback channel not in sync with normal backend channel which might cause conflicts when you shutdown streaming application, not sure it worth the effort at this point (also discussed with Daniel offline), let's ask the question to pyspark guys how they handle it.

xiongrenyi: We need better synchronization in sparkclr which means serializing calls to jvm l believe so that we can follow the same model by calling ssc.AwaitTermination instead

PR #54, opened 2015-11-14
auto copy dependent jars to run\lib; auto fill SPARKCSV_JARS

danielli90:
This PR removes two manual steps before running samples locally (1) manually download/copy dependent jars; (2) manually set SPARKCVS_JARS to include all dependent jars.


jthelin:
Nice! The more friction that can be removed from the process of getting new people able to try something, the better all round.

danielli90:
@qintao1976 suggested to create a uber package using maven. Will evaluate - possibly create a new pr if uber package works out.

danielli90:
replaced by another PR which builds uber jar

PR #55, opened 2015-11-15
Create a uber package to include all dependency

danielli90:
it increases the target sample jar from 154KBytes to 104MBytes, but it **removes** the error-prone manual processes in manually identify dependent  jars needed at run-time, and manually include those jars at SPARKCVS_JAR before submitting the job. this PR replaces #54 - which will not be merged in.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -35,6 +35,7 @@ csharp/Samples/Microsoft.Spark.CSharp/obj/**
 csharp/Worker/Microsoft.Spark.CSharp/bin/**
 csharp/Worker/Microsoft.Spark.CSharp/obj/**
 scala/.idea/**
+scala/dependency-reduced-pom.xml
 run/
 *.log
 lib/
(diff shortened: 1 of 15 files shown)


danielli90:
replaced by another PR that squashes two commits into one

jthelin:
BTW @danielli90 you don't have to create a new PR to do this type of "branch cleanup".

You can rebase/squash your local copy of the branch to whatever "final form" you want and then (force) push to the same branch name to update the original PR on GitHub.

That will just replace the change set to be merged for this PR but keep all the PR discussion history.

Most repos have their `master` branch set to prohibit forced pushes on that branch [and rightly so, IMHO -- should always merge to master, not rewrite] 
But it actually makes a lot of sense to do forced pushes on some long-running pull request / feature branches where lots of discussion & changes happened before the final stable code version is ready to merge in to master.

Many people have an automatic aversion to `git push --force` but it is actually a very useful function when used for the correct reasons. This is one of those "good" use cases ;)
HTH.

danielli90:
Got it. Thanks @jthelin.

danielli90:
Used @jthelin’s suggestion for PR#59 – “git push –force” worked magically.

Thx
Daniel

PR #59, opened 2015-11-17
auto download missing build tools and local run-time dependencies

danielli90:


danielli90:
Auto download build tools (nuget, maven) and run-time dependencies (apache-spark, winutils) and set up environment variables for build and runsamples. Also check whether developer command prompts are used and whether JAVA_HOME is set properly.

downloadtools.ps1
```diff
@@ -0,0 +1,302 @@
...
+    }
+    
+    if (!($path -like "*$toolsDir*"))
+    {
+        # add toolsdir to path
+        $envStream.WriteLine("set path=$toolsDir;%path%");
+    }
+
+    # Download Spark binary
+    $S_HOME = "$toolsDir\spark-1.4.1-bin-hadoop2.6"
```
jthelin: Minor nit: you might want to pull out variables for spark-version and hadoop-version then substitute them into the various places in theses strings. That way, only one line change to change version :)

jthelin:
This is a great change! Really helps remove friction for new users in the getting-started stage :)

xiongrenyi:
looks good to me

PR #61, opened 2015-11-17
RunSamples.cmd could receive arguments

tawan0109:
1. Extra arguments can be passed to RunSamples.cmd.
2. Now single sample can be launched by 'RunSamples.cmd sparkclr.samples.torun <SampleName>'.
3. Options that is supported by SparkCLRSamples.exe can also be supported by RunSamples.cmd automatically.

### DIFF
--- a/RunSamples.cmd
+++ b/RunSamples.cmd
@@ -22,4 +22,4 @@ cd %SPARKCLR_HOME%
 
 @echo ON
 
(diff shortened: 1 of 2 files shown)


danielli90:
Very promising and exciting improvement, daily.

Sent from my Windows Phone
________________________________

PR #63, opened 2015-11-17
Dev2

xiongrenyi:


xiongrenyi:
should have pull from master instead of dev2 (mistake)

PR #65, opened 2015-11-18
After RunSamples.cmd execution finishes, navigate back to the original directory, instead of staying in '%SPARKCLR_HOME%'

tawan0109:


danielli90:
LGTM (looks good to me)

PR #68, opened 2015-11-19
Microsoft.VisualStudio.QualityTools.UnitTestFramework not found at runtime

danielli90:
Microsoft.VisualStudio.QualityTools.UnitTestFramework is not part of .Net Framework (just part of VS IDE); by enabling "Copy Local" to true in the reference, the DLL will be included as part of final build, which avoids the exception of missing UnitTestFramework.dll when running on servers without VS IDE installed.

Running "sparkclr.samples.torun PiSample" in the cluster without VS IDE installed repro's this issue.

### DIFF
--- a/csharp/Samples/Microsoft.Spark.CSharp/Samples.csproj
+++ b/csharp/Samples/Microsoft.Spark.CSharp/Samples.csproj
@@ -33,7 +33,9 @@
     <WarningLevel>4</WarningLevel>
   </PropertyGroup>
   <ItemGroup>
-    <Reference Include="Microsoft.VisualStudio.QualityTools.UnitTestFramework" />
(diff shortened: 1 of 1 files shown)


jthelin:
The other way to deal with this type of problem is to switch to using a different "Assertions" package from NuGet rather than `Microsoft.VisualStudio.QualityTools.UnitTestFramework`.

`FluentAssertions` is one I have used, but there are other examples like `NFluent` etc. too.
http://www.fluentassertions.com/
http://n-fluent.net/

I personally like the "fluent" style of writing assertion because it is so much easier to read and get right.

With the old-school Assert.Blah(..) methods I often have to keep checking which way round the paramaters go, and it is different for AreEquals and IsInstanceOfType for example. :(

danielli90:
Thanks @jthelin. Created #75 to track this suggestion.

PR #77, opened 2015-11-20
Promote some widely used string variables to const values for easy maintenance purpose & Fix build warning

tawan0109:
* Use `public const string` to declare those widely referenced string values for easy maintenance purpose.
* Fix build warning.


csharp/Adapter/Microsoft.Spark.CSharp/Configuration/ConfigurationService.cs
```diff
@@ -79,7 +83,6 @@ private class SparkCLRConfiguration
             protected readonly AppSettingsSection appSettings;
             protected readonly string sparkCLRHome = Environment.GetEnvironmentVariable("SPARKCLR_HOME"); //set by sparkclr-submit.cmd
```
jthelin: Nit: "SPARKCLR_HOME" is probably another string that is const-able.

PR #85, opened 2015-11-24
Build.cmd does not stop when build errors happen

tawan0109:
* When error happens during `scala` or `csharp` components building, build.cmd does not stop and continue to do the left steps.
* Check `ERRORLEVEL` and stop building if the `ERRORLEVEL` is not equal `0`

### DIFF
--- a/Build.cmd
+++ b/Build.cmd
@@ -43,6 +43,11 @@ call mvn.cmd clean
 @rem build the package
 call mvn.cmd package
 
+if %ERRORLEVEL% NEQ 0 (
+	@echo Build SparkCLR Scala components failed, stop building.
+	popd
+	goto :eof
+)
 @echo SparkCLR Scala binaries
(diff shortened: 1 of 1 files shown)


danielli90:
LGTM.

Build.cmd
```diff
@@ -43,6 +43,11 @@ call mvn.cmd clean
 @rem build the package
 call mvn.cmd package
 
+if %ERRORLEVEL% NEQ 0 (
+	@echo Build SparkCLR Scala components failed, stop building.
+	popd
+	goto :eof
```
jthelin: I think you also want to ensure than this script [here and below] returns a non-zero result code if a build step fails. 

Compare with the `goto :ErrorStop` and `exit /b %RC%` code in the "other" Build.cmd script.
https://github.com/Microsoft/SparkCLR/blob/master/csharp/Build.cmd#L67

tawan0109: Good point, I will send another PR for it.

Issue #111, opened 2015-12-02
Inconsistent addressing schemes used by sockets in SparkCLR


kai-zeng:
In Microsoft.Spark.CSharp.Core.RDD, method Collect creates a socket without explicitly specifying the AddressFamily, which by default is InterNetworkv6 (ip v6). This also cause "invalid argument" exceptions on Linux when calling collect on RDD.
This is also inconsistent with other places in SpakCLR, e.g., SparkCLRSocket or JvmBridge, where sockets always explicitly use AddressFamily.InterNetwork (ip v4).

PR #112, opened 2015-12-02
[Issue #111]Fix Inconsistent addressing schemes used by sockets in SparkCLR

kai-zeng:
In Microsoft.Spark.CSharp.Core.RDD, method Collect creates a socket without explicitly specifying the AddressFamily, which by default is InterNetworkv6 (ip v6). This also cause "invalid argument" exceptions on Linux when calling collect on RDD.
This is also inconsistent with other places in SpakCLR, e.g., SparkCLRSocket or JvmBridge, where sockets always explicitly use AddressFamily.InterNetwork (ip v4).

### DIFF
--- a/csharp/Adapter/Microsoft.Spark.CSharp/Core/RDD.cs
+++ b/csharp/Adapter/Microsoft.Spark.CSharp/Core/RDD.cs
@@ -554,7 +554,7 @@ namespace Microsoft.Spark.CSharp.Core
         internal IEnumerable<dynamic> Collect(int port)
         {
             IFormatter formatter = new BinaryFormatter();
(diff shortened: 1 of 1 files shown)


skaarthik:
LGTM

danielli90:
LGTM

xiongrenyi:
Yes, should get consistent. Will fix it.

Curious, did you actually observe any problems?

Since client side should be overridden by address family used in s.Connect(“127.0.0.1”, port)

Thanks,
Renyi.

qintao1976:
This PR looks fine for me.

kai-zeng:
@xiongrenyi Yes, I got SocketException saying "Invalid Argument" when running using Mono on Linux. On windows it works fine though.

PR #115, opened 2015-12-02
add runsamples; proceed with build even not dev command prompt for VS

danielli90:
This change added complete sample run, with --validate flag, after unit test. It increases build-time at AppVeyor from 5 minute to 13 minutes and 30 seconds. We will observe closely whether the increase in build-time negatively impact the productivity - if it does, we will run selected sample test.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -40,3 +40,4 @@ run/
 tools/
 *.log
 lib/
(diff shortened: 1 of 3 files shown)


qintao1976:
LGTM

PR #118, opened 2015-12-03
workaround to build uber pacakge in build.cmd using shade-plugin

danielli90:
Without breaking debug mode in IntelliJ


qintao1976:
LGTM

PR #119, opened 2015-12-03
Enable scala style check in pom.xml

tawan0109:
* Enable scala style check in `pom.xml`
* Use same `scalastyle-config.xml` with spark.
* Set `failOnViolation` to `false` for `scalastyle-maven-plugin` for now.
* Use `mvn verify` to check scala style, results can be found in `target/scalastyle-output.xml`
* If this PR accepted, will make changes to scala files to resolve detected warnings.

### DIFF
--- a/scala/scalastyle-config.xml
+++ b/scala/scalastyle-config.xml
@@ -37,6 +37,10 @@ This file is divided into 3 sections:
  (3) rules that we don't want to enforce.
 -->
 
+<!--
+Reference: Spark scalastyle-config.xml (https://github.com/apache/spark/blob/master/scalastyle-config.xml)
+-->
+
 <scalastyle>
   <name>Scalastyle standard configuration</name>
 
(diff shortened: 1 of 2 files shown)


scala/scalastyle-config.xml
```diff
@@ -0,0 +1,248 @@
...
+  <!-- ================================================================================ -->
+  <!--                               rules we enforce                                   -->
+  <!-- ================================================================================ -->
+
+  <check level="error" class="org.scalastyle.file.FileTabChecker" enabled="true"></check>
+
+  <check level="error" class="org.scalastyle.file.HeaderMatchesChecker" enabled="true">
+    <parameters>
+       <parameter name="header"><![CDATA[/*
+* Copyright (c) Microsoft. All rights reserved.
```
danielli90: Since this file is from https://github.com/apache/spark/blob/master/scalastyle-config.xml, I think we can just an URL reference here, and remove MIT license text for now.

danielli90:
Added one comment above - suggesting adding a reference to https://github.com/apache/spark/blob/master/scalastyle-config.xml. Otherwise, LGTM.

scala/scalastyle-config.xml
tawan0109: Reference is added in a new commit.
MIT license text here is used to check the header of each scala source file in SparkCLR, not a license.

PR #127, opened 2015-12-04
enabled "provided" flag in pom.xml, when creating uber package

danielli90:


tawan0109:
LGTM

skaarthik:
It is not a good idea to keep deviating from the behavior one would see when running "mvn package". With the behind the scenes updates made through powershell scripts we are implicitly enforcing one way of building the code that would result in a behavior not consistent with what one would expect from the POM file that is in the repo. I think we should move away from doing anything silently at the time of build and make everything that happens in build, packaging and running samples explicit and straightforward.

danielli90:
Agreed. Created #130 to track this to-do item.

PR #136, opened 2015-12-06
Update to nunit3 as appveyor is upgraded

kai-zeng:
Update unit test to use NUnit3, as AppVeyor is upgraded to use NUnit3.
Explicitly set assemblies to test, as AppVeyor test auto-detection will detect SparkCLRSamples.exe but fail to find any testfixtures in it.


skaarthik:
LGTM

rapoth:
LGTM

PR #137, opened 2015-12-07
Revert "Update to nunit3 as appveyor is upgraded"

rapoth:
Reverts Microsoft/SparkCLR#136


rapoth:
Looks like the previous PR has some problems. Will revert until that is fixed.

PR #145, opened 2015-12-08
added back AssemblyInfo.cs for version stamps

danielli90:
to prepare for the release later

### DIFF
--- a/csharp/Adapter/Microsoft.Spark.CSharp/Properties/AssemblyInfo.cs
+++ b/csharp/Adapter/Microsoft.Spark.CSharp/Properties/AssemblyInfo.cs
@@ -1,0 +1,34 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+using System;
+using System.Reflection;
+using System.Runtime.CompilerServices;
+using System.Runtime.InteropServices;
+
+// General Information about an assembly is controlled through the following 
+// set of attributes. Change these attribute values to modify the information
+// associated with an assembly.
+[assembly: AssemblyTitle("Microsoft.Spark.CSharp.Adapter")]
+[assembly: AssemblyDescription("Adapter provides Spark public APIs for C# driver program.")]
+[assembly: AssemblyConfiguration("")]
+[assembly: AssemblyCompany("Microsoft Corporation")]
+[assembly: AssemblyProduct("Microsoft SparkCLR")]
+[assembly: AssemblyCopyright("Copyright © Microsoft Corporation 2015")]
+[assembly: AssemblyTrademark("")]
+[assembly: AssemblyCulture("")]
+
+// Setting ComVisible to false makes the types in this assembly not visible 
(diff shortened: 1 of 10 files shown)


qintao1976:
LGTM

scala/pom.xml
```diff
@@ -2,7 +2,7 @@
   <modelVersion>4.0.0</modelVersion>
   <groupId>com.microsoft.spark</groupId>
   <artifactId>spark-clr</artifactId>
-  <version>${spark.version}-SNAPSHOT</version>
+  <version>1.4.1-SNAPSHOT</version>
```
skaarthik: why do you want to make this change?

csharp/Worker/Microsoft.Spark.CSharp/Properties/AssemblyInfo.cs
```diff
@@ -0,0 +1,32 @@
+﻿using System;
```
skaarthik: Copyright & License info missing

csharp/Adapter/Microsoft.Spark.CSharp/Properties/AssemblyInfo.cs
```diff
@@ -0,0 +1,32 @@
+﻿using System;
+using System.Reflection;
+using System.Runtime.CompilerServices;
+using System.Runtime.InteropServices;
+
+// General Information about an assembly is controlled through the following 
+// set of attributes. Change these attribute values to modify the information
+// associated with an assembly.
+[assembly: AssemblyTitle("Microsoft.Spark.CSharp.Adapter")]
+[assembly: AssemblyDescription("")]
```
skaarthik: Do we want to make use of this tag and provide a short description to the 3 binaries we release?

scala/pom.xml
danielli90: mvn has warnings for this macro

csharp/Worker/Microsoft.Spark.CSharp/Properties/AssemblyInfo.cs
danielli90: Added.

csharp/Adapter/Microsoft.Spark.CSharp/Properties/AssemblyInfo.cs
danielli90: Added description.

scala/pom.xml
skaarthik: What is the warning? If it is not serious we may want to keep the macro as it helps keeping version in one place in the file.

danielli90: Copied warning below. Since we will map each branch to a specific spark version at a given time - a constant spark-version in pom.xml is still managable.

[WARNING]
[WARNING] Some problems were encountered while building the effective model for com.microsoft.spark:spark-clr:jar:1.4.1-SNAPSHOT
[WARNING] 'version' contains an expression but should be a constant. @ com.microsoft.spark:spark-clr:${spark.version}-SNAPSHOT, c:\gitsrc\SparkCLR\scala\pom.xml, line 5, column 12

Issue #84, opened 2015-11-24
Define a public constant for SPARKCLR_HOME in ConfigurationService.cs


danielli90:
"SPARKCLR_HOME" should be defined as a public constant in ConfigurationService.cs. See original comment in #77

jthelin:
I think this is now fixed by PR #168

qintao1976:
yes, it is fixed by PR #168.
I close the issue.
Thanks

PR #168, opened 2015-12-16
 Define a public constant for SPARKCLR_HOME in ConfigurationService.cs #84 

ms-guizha:


ms-guizha:
I have sent codereview  using codeflow and tao has signed off for me, before sending out his PR, I have run unittest using VS and run samples

PR #169, opened 2015-12-17
spark 1.5.2 upgrade

skaarthik:


csharp/Adapter/Microsoft.Spark.CSharp/Sql/DataFrame.cs
```diff
@@ -26,14 +24,14 @@ public class DataFrame
...
         [NonSerialized]
         private RDD<Row> rdd;
-
+        [NonSerialized]
+        private IRDDProxy rddProxy;
         [NonSerialized]
         private bool? isLocal;
         [NonSerialized]
-        private Random random = new Random();
+        private readonly Random random = new Random();
```
jthelin: Not directly related to to this change, but it is very important to remember that the `Random` class is NOT thread-safe.
https://msdn.microsoft.com/en-us/library/system.random(v=vs.110).aspx#ThreadSafety

To make things even worse, the side effect of multi-threaded usage of an instance of `Random` class is a massively degenerate failure mode -- it starts to always return `0` on each call ! -- ie, about as skewed and far away from "random" as you could possibly get!

We got hit by that really nasty "feature" of `Random` in Orleans project, and took a long time to work out what was going wrong. 
I fear that client code passing around `DataFrame` object references could easily trigger this same "feature". :(

In Orleans we ended up creating a `SafeRandom` wrapper as a drop-in replacement, which i would be happy to send across as a separate PR if you want?
https://github.com/dotnet/orleans/blob/a64fa8953e764ac391ebd549d395ac66374b545f/src/Orleans/Utils/SafeRandom.cs

csharp/AdapterTest/AdapterTest.csproj
```diff
@@ -94,9 +96,7 @@
       <Name>Adapter</Name>
     </ProjectReference>
   </ItemGroup>
-  <ItemGroup>
-    <Folder Include="Properties\" />
-  </ItemGroup>
+  <ItemGroup />
```
jthelin: Nit: unused XML element

csharp/Adapter/Microsoft.Spark.CSharp/Adapter.csproj
```diff
@@ -121,9 +125,7 @@
     <Compile Include="Streaming\StreamingContext.cs" />
     <Compile Include="Streaming\TransformedDStream.cs" />
   </ItemGroup>
-  <ItemGroup>
-    <Folder Include="Properties\" />
-  </ItemGroup>
+  <ItemGroup />
```
jthelin: Nit: Empty XML element.

csharp/Adapter/Microsoft.Spark.CSharp/Sql/Row.cs
```diff
@@ -235,7 +247,7 @@ internal RowImpl(object data, RowSchema schema)
             }
 
             this.schema = schema;
-
+
```
jthelin: Nit: extra spaces introduced here.

csharp/AdapterTest/DataFrameTest.cs
```diff
@@ -643,70 +647,88 @@ public void TestLimit()
...
+            mockDataFrameProxy.Setup(m => m.Limit(It.IsAny<int>())).Returns(mockDataFrameProxy.Object);
+            var dataFrame = new DataFrame(mockDataFrameProxy.Object, null);
+
+            const int unUsedSizeValue = 100;
 
             // act
-            IEnumerable<Row> rows = dataFrame.Head(size);
-
+            IEnumerable<Row> rows = dataFrame.Head(unUsedSizeValue); //uses limit & collect
+
```
jthelin: Nit: extra spaces introduced here.

csharp/Worker/Microsoft.Spark.CSharp/Worker.cs
```diff
@@ -425,7 +425,9 @@ private object[] GetNext(int messageLength)
                     {
                         Debug.Assert(messageLength > 0);
                         byte[] buffer = SerDe.ReadBytes(inputStream, messageLength);
-                        result = unpickler.loads(buffer) as object[];
+                        var unpickledObjects = PythonSerDe.GetUnpickledObjects(buffer);
```
jthelin: Seems that a new Unpickler instance will get create by PythonSerDe.GetUnpickledObjects for every row received? 

If i understand this code path correctly, does the Unpicker instance need to live on outside this method for the whole lifetime of the Row being created, or is it only needed at this point of Row creation?

In other words, can Unpicker instances be cached & re-used for different Rows [maybe ThreadLocal?] to avoid creating so many new objects?

csharp/Adapter/Microsoft.Spark.CSharp/Sql/DataFrame.cs
skaarthik: Our usage of random is pretty trivial and just in one non-critical method which is not multi-threaded. I believe we do not need SafeRandom right away but thanks for the info and we should keep an eye on the bugs in Sample method to see if it ever starts exhibiting the "feature" you described.

csharp/AdapterTest/AdapterTest.csproj
skaarthik: Will address the Nits you pointed out in a different PR

csharp/Worker/Microsoft.Spark.CSharp/Worker.cs
skaarthik: I do not know the recommended usage for Unpickler and hence going with the current implementation. Pickling and unpickling is no longer a black box with the update to 1.5.2. So it is important to understand what is under the hood and make necessary perf tuning. That is one of the top items in my todo list. Hopefully I can provide a better response to your comment as I get started on that.

jthelin: Sounds good. Cheers.

kai-zeng:
close and reopen to trigger a new build

PR #183, opened 2015-12-22
Use Path.Combine to building worker exe path.

jthelin:
- Use Path.Combine for better cross-platform support for building the worker exe path, rather than hard coding the Windows specific backslash path separator.

- Assembly.Location can return null in some cases (particularly dynamic loading) so default to current directory "." in that case.

- Output the worker exe path being used, for easier debugging of tests not working.

- Use object initializer syntax for creating Process.StartInfo -- more compact and readable code.

### DIFF
--- a/csharp/Samples/Microsoft.Spark.CSharp/MiscSamples.cs
+++ b/csharp/Samples/Microsoft.Spark.CSharp/MiscSamples.cs
@@ -19,7 +19,7 @@ namespace Microsoft.Spark.CSharp.Samples
         [Sample]
         internal static void PiSample()
         {
-            const int slices = 2;
+            const int slices = 3;
             var n = (int) Math.Min(100000L*slices, int.MaxValue);
             var values = new List<int>(n);
             for (int i = 0; i <= n; i++)
--- a/csharp/WorkerTest/WorkerTest.cs
+++ b/csharp/WorkerTest/WorkerTest.cs
@@ -36,11 +36,18 @@ namespace WorkerTest
             tcpListener.Start();
             int port = (tcpListener.LocalEndpoint as IPEndPoint).Port;
 
-            worker = new Process();
-            worker.StartInfo.FileName = Path.GetDirectoryName(Assembly.GetExecutingAssembly().Location) + "\\CSharpWorker.exe";
-            worker.StartInfo.UseShellExecute = false;
-            worker.StartInfo.RedirectStandardInput = true;
-            worker.StartInfo.RedirectStandardOutput = true;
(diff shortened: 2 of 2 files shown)


xiongrenyi:
LGTM

jthelin:
Travis build fails with error below, but that looks like it is just a limit precision problem in the  `PiSample` test case itself:

3.16194 - 3.14 = 0.02194 rather than <= 0.019 expected.
``` c#
Assert.IsTrue(Math.Abs(approximatePiValue - 3.14) <= 0.019);
```

I suspect setting `slices = 3` (currently 2) in  `PiSample` will fix this, although not directly connected to the change in this PR.

```
----- Running sample PiSample -----
Pi is roughly 3.13796
Pi is roughly 3.16194
----- Error running sample PiSample (duration=00:00:11.2844210) -----
System.Reflection.TargetInvocationException: Exception has been thrown by the target of an invocation. 
---> NUnit.Framework.AssertionException:   
  Expected: True
  But was:  False
  at NUnit.Framework.Assert.That[TActual] (NUnit.Framework.TActual actual, IResolveConstraint expression, System.String message, System.Object[] args) <0x420b4670 + 0x000fb> in <filename unknown>:0 
  at NUnit.Framework.Assert.IsTrue (Boolean condition) <0x420b4540 + 0x0001f> in <filename unknown>:0 
  at Microsoft.Spark.CSharp.Samples.MiscSamples.PiSample () <0x420b3f70 + 0x0044f> in <filename unknown>:0 
  at (wrapper managed-to-native) System.Reflection.MonoMethod:InternalInvoke (System.Reflection.MonoMethod,object,object[],System.Exception&)
  at System.Reflection.MonoMethod.Invoke (System.Object obj, BindingFlags invokeAttr, System.Reflection.Binder binder, System.Object[] parameters, System.Globalization.CultureInfo culture) <0x7f446d5ab780 + 0x000a1> in <filename unknown>:0 
  --- End of inner exception stack trace ---
  at System.Reflection.MonoMethod.Invoke (System.Object obj, BindingFlags invokeAttr, System.Reflection.Binder binder, System.Object[] parameters, System.Globalization.CultureInfo culture) <0x7f446d5ab780 + 0x000ef> in <filename unknown>:0 
  at System.Reflection.MethodBase.Invoke (System.Object obj, System.Object[] parameters) <0x7f446d3e3cf0 + 0x0002a> in <filename unknown>:0 
  at Microsoft.Spark.CSharp.Samples.SamplesRunner.RunSample (System.Reflection.MethodInfo sample) <0x41ffddd0 + 0x0036c> in <filename unknown>:0 
```

PR #184, opened 2015-12-22
include dependent packages, symbols and config files

danielli90:
This is to ensure nuget package can be used to compile C# drive program. Tested with all samples.


tawan0109:
LGTM

PR #185, opened 2015-12-22
load usage help text for samples from file

ms-guizha:


danielli90:
LGTM.

csharp/Samples/Microsoft.Spark.CSharp/samplesusage.md
```diff
@@ -0,0 +1,50 @@
+SparkCLRSamples.exe supports following options:
+
```
skaarthik: LGTM

Consider adding bullets to the options and also format the arg names and values (for example, argument name like --temp or spark.local.dir could be in bold and value like <TEMP_DIR> could be in italics)

danielli90: Because the same text file needs to be displayed at console, we may just live with this simple format for now. Otherwise, we need to add logic to remove markdown before displaying at console - which doesn't seem worth the effort now.

Sent from my Windows Phone
________________________________

skaarthik:
@ms-guizha - @danielli90 made a valid point on formatting. Ignore my previous comment.

csharp/Samples/Microsoft.Spark.CSharp/samplesusage.md
ms-guizha: Ok, I get your point

PR #186, opened 2015-12-22
convert PiSample to PiExample consuming SparkCLR nuget package

danielli90:
A developer should refer SparkCLR nuget package, when writing a SparkCLR job. This is the first attempt in offering a PiExample. Borrowed all code from existing SparkCLR sample run.

Tested by including SparkCLR package from local folder.

Next: need to provide a readme to cover (1) how to build; (2) how to submit the job.

### DIFF
--- a/examples/Examples.sln
+++ b/examples/Examples.sln
@@ -1,0 +1,20 @@
+﻿
+Microsoft Visual Studio Solution File, Format Version 12.00
+# Visual Studio 2013
+Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Pi", "Pi\Pi.csproj", "{913E6A56-9839-4379-8B3C-855BA9341663}"
+EndProject
+Global
+	GlobalSection(SolutionConfigurationPlatforms) = preSolution
+		Debug|Any CPU = Debug|Any CPU
+		Release|Any CPU = Release|Any CPU
+	EndGlobalSection
+	GlobalSection(ProjectConfigurationPlatforms) = postSolution
+		{913E6A56-9839-4379-8B3C-855BA9341663}.Debug|Any CPU.ActiveCfg = Debug|Any CPU
+		{913E6A56-9839-4379-8B3C-855BA9341663}.Debug|Any CPU.Build.0 = Debug|Any CPU
+		{913E6A56-9839-4379-8B3C-855BA9341663}.Release|Any CPU.ActiveCfg = Release|Any CPU
+		{913E6A56-9839-4379-8B3C-855BA9341663}.Release|Any CPU.Build.0 = Release|Any CPU
+	EndGlobalSection
+	GlobalSection(SolutionProperties) = preSolution
+		HideSolutionNode = FALSE
+	EndGlobalSection
+EndGlobal
--- a/examples/pi/App.config
+++ b/examples/pi/App.config
@@ -1,0 +1,39 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<configuration>
+  <!--*************************************************************************************-->
+  <!-- Log4Net is used by the example -->
+  <!--**************************************************************************************-->
+  <configSections>
+    <section name="log4net" type="log4net.Config.Log4NetConfigurationSectionHandler, log4net" />
+  </configSections>
+
+ <log4net>
+   <root>
+     <level value="DEBUG" />
+       <appender-ref ref="ConsoleAppender" />
+       <!--<appender-ref ref="LogFileAppender" />-->
+   </root>
+
+    <appender name="ConsoleAppender" type="log4net.Appender.ConsoleAppender">
+      <layout type="log4net.Layout.PatternLayout">
+        <conversionPattern value="[%date] [%thread] [%-5level] [%logger] - %message%newline" />
+      </layout>
+    </appender>
+
+    <appender name="LogFileAppender" type="log4net.Appender.RollingFileAppender">
+      <file type="log4net.Util.PatternString">
+        <conversionPattern value="%env{TEMP}\\SparkCLRLogs\\SparkCLR-Pi_%env{COMPUTERNAME}[%processid].log" />
+      </file>
+      <param name="AppendToFile" value="true" />
+      <param name="MaxSizeRollBackups" value="2000" />
+      <param name="MaxFileSize" value="51200000" />
+      <param name="StaticLogFileName" value="false" />
+      <param name="DatePattern" value=".yyyy_MM_dd_hh" />
+      <param name="RollingStyle" value="Composite" />
+      <layout type="log4net.Layout.PatternLayout">
+        <conversionPattern value="[%date] [%thread] [%-5level] [%logger] - %message%newline" />
+      </layout>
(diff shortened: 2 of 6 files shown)


tawan0109:
LGTM.

skaarthik:
Having two folders examples and samples under csharp folder could be confusing. We might want to rename the new folder to NuGetSample or something that reflects its purpose. Also, I recommend keeping it simple without all the bells and whistles in samples project because it is just used for illustrating the usage of NuGet sample. Ideally users should be able just copy that project and start adding their code. So I recommend keeping it simple with nothing more than creating the context and calculating Pi - everything else like command line parsing etc. could be removed.

danielli90:
Similar to Apache Spark, "examples" folder is top-level folder, serving as an invitation for developers who want to write C# drive program (but not into contributing to the core of SparkCLR). "samples" folder is under "csharp" - with its current purpose focusing on validating public API supported SparkCLR, tailored to developers who are contributing to SparkCLR core. Placing these two folders under different directories, hopefully to differentiate and avoid confusion. 

The reason to have a structure for running multiple examples, is to model - https://github.com/apache/spark/blob/master/bin/run-example, which allows multiple examples to be launched.

We definitely need to find options to converge "examples" and "samples", to avoid duplication down the road.

Hope this makes sense.

examples/Examples.csproj
```diff
@@ -0,0 +1,103 @@
...
+    </Reference>
+    <Reference Include="nunit.framework, Version=3.0.5813.39031, Culture=neutral, PublicKeyToken=2638cd05610744eb, processorArchitecture=MSIL">
+      <SpecificVersion>False</SpecificVersion>
+      <HintPath>packages\NUnit.3.0.1\lib\net45\nunit.framework.dll</HintPath>
+    </Reference>
+    <Reference Include="CSharpWorker, Version=1.5.2.0, Culture=neutral, processorArchitecture=MSIL">
+      <SpecificVersion>False</SpecificVersion>
+      <HintPath>packages\Microsoft.SparkCLR.1.5.2-SNAPSHOT\lib\net45\CSharpWorker.exe</HintPath>
+    </Reference>
+    <Reference Include="log4net, Version=1.2.15.0, Culture=neutral, PublicKeyToken=669e0ddf0bb1aa2a, processorArchitecture=MSIL">
```
jthelin: Using full assembly names here is the reason why the bindingRedirects were needed in App.config.

See more background to this problem in the "External Assembly References" section notes of Project Orleans.
http://dotnet.github.io/orleans/Contributing

examples/MiscExamples.cs
```diff
@@ -0,0 +1,71 @@
...
+{
+    class MiscExamples
+    {
+        /// <summary>
+        /// Example to compute th evalue of Pi
+        /// </summary>
+        [Example]
+        internal static void PiExample()
+        {
+            const int slices = 2;
```
jthelin: This will likely suffer from the same test precision problems that PR #183 hit on Travis CI build.

Suggest using `slices = 3` might be less brittle.

examples/Examples.csproj
```diff
@@ -0,0 +1,103 @@
...
+    <DebugType>pdbonly</DebugType>
+    <Optimize>true</Optimize>
+    <OutputPath>bin\Release\</OutputPath>
+    <DefineConstants>TRACE</DefineConstants>
+    <ErrorReport>prompt</ErrorReport>
+    <WarningLevel>4</WarningLevel>
+  </PropertyGroup>
+  <ItemGroup>
+    <Reference Include="Newtonsoft.Json, Version=7.0.0.0, Culture=neutral, PublicKeyToken=30ad4fe6b2a6aeed, processorArchitecture=MSIL">
+      <SpecificVersion>False</SpecificVersion>
```
jthelin: NuGet tool logic fail! 

Nuget uses specific fully-qualified assembly name here (including a 4-digit specific assembly version), then says "SpecificVersion=False".

You have to LOL or else you would cry! :(

examples/App.config
```diff
@@ -0,0 +1,59 @@
...
+    <!--********************************************************************************************************-->
+    <!--** Uncomment the following settings to run Spark driver executable in **local** or **debug** modes ** -->
+    <!--** In debug mode, the driver is not launched by CSharpRunner but launched from VS or command prompt not configured for SparkCLR ** -->
+    <!--** CSharpBackend should be launched in debug mode as well and the port number from that should be used below ** -->
+    <!--** Command to launch CSharpBackend in debug mode is "sparkclr-submit.cmd debug" ** -->
+    <!--********************************************************************************************************-->
+    
+    <!--
+    <add key="CSharpWorkerPath" value="C:\SparkCLR\csharp\Examples\bin\Debug\CSharpWorker.exe"/>    
+    <add key="CSharpBackendPortNumber" value="0"/>
```
jthelin: Is this stuff needed? 

I suggest might be clearer if Examples code targets what "normal" end-user developer will do, rather than any special debug that core SparkCLR dev team might do.

Goal for Examples should really be: "Just works by pressing `F5` in Visual Studio"

examples/Properties/AssemblyInfo.cs
```diff
@@ -0,0 +1,34 @@
...
+using System;
+using System.Reflection;
+using System.Runtime.CompilerServices;
+using System.Runtime.InteropServices;
+
+// General Information about an assembly is controlled through the following 
+// set of attributes. Change these attribute values to modify the information
+// associated with an assembly.
+[assembly: AssemblyTitle("SparkCLRExamples")]
+[assembly: AssemblyDescription("Utility for running all or selected SparkCLR samples.")]
```
jthelin: Nit: s/samples/examples/

skaarthik:
Sorry, I am still not convinced. Samples project serves two purposes - usage samples for the public API (for users) and leveraging the same as functional test suite (for contributors) to avoid duplication of code. If a developer wants to contribute a sample/example it could be done to the samples project with or without validation enabled. This will help keeping all the samples/examples in one code and avoid discoverability or code duplication problems.

The only additional value this project brings in is on using nuget package and a template project for starters. I do not want to see this new project become redundant to the samples project.

danielli90:
Ok. Will reduce PiExample to one simple program to remove redundancy.

skaarthik:
Thanks @danielli90

examples/Examples.csproj
danielli90: Thanks @jthelin. Fixed.

examples/MiscExamples.cs
danielli90: fixed

examples/Examples.csproj
danielli90: fixed

examples/App.config
danielli90: removed. Not as simple as F5 though - will provide a readme.md on how to compile and deploy locally and in a cluster.

examples/Properties/AssemblyInfo.cs
danielli90: fixed. Reduced the example code to Pi calculation.

skaarthik:
LGTM

PR #187, opened 2015-12-22
including workertest in code coverage

skaarthik:


skaarthik:
Only config change and no code changes. Config change validated using the code coverage results published in CodeCov site. So merging the PR without LGTM from other committers.

PR #192, opened 2015-12-23
Show progress bar when downloading runtime dependencies

tawan0109:
When execute `RunSamples.cmd` for the first time on Windows, `RunSamples.cmd` will try to download all runtime dependencies. For dependencies like `spark-1.5.2-bin-hadoop2.6.tgz`, the file size is kind of large.

During downloading, for now user can't know how is the downloading going. Sometimes user might get confused as there is no response before downloading completes(especially when the internet connection is not good enough).

This PR tries to show progress bar when downloading files. 

Preview


skaarthik:
Nice!

LGTM

danielli90:
ditto here. LGTM.

PR #193, opened 2015-12-23
Minor change to class `StructType` to make it easy to test

tawan0109:
As interface `IStructTypeProxy` has already defined method `ToJson()`, so here no need to cast `structTypeProxy ` to class `StructTypeIpcProxy`. This change will make class `StructType` easy to test.


jayjaywg:
LGTM

PR #197, opened 2016-01-01
Add profile "uber-jar" in pom file to create uber package

qintao1976:


danielli90:
Just realized that the uber package size increased from about 30MBytes to 100MBytes, due to the removal of "provided" tag, as in PR#127

I will include re-enabling "provided" tag in another PR related to package signing.

Discussed with @skaarthik, eventually in M2, we prefer to avoid uber package.

PR #203, opened 2016-01-05
moving streaming methods from top level proxy and other minor updates

skaarthik:


danielli90:
LGTM with comments below.

corefx coding guideline (https://github.com/dotnet/corefx/blob/master/Documentation/coding-guidelines/coding-style.md) explicitly called out "We avoid  this.  unless absolutely necessary. ". However, StyleCop is in favor of the "this." prefix - (1) http://www.stylecop.com/docs/SA1309.html; (2) http://www.stylecop.com/docs/SA1101.html. This change indicates that we will choose bullet-point-4 corefx coding-style.md - which is fine with me. This can be the starting point when we phase in coding style for C#.

Specific to change in line#42 in PipelinedRDD.cs, "this." prefix seems to me actually helping readability though. Added inline comment as well.

csharp/Adapter/Microsoft.Spark.CSharp/Core/PipelinedRDD.cs
```diff
@@ -39,13 +39,12 @@ public override RDD<U1> MapPartitionsWithIndex<U1>(Func<int, IEnumerable<U>, IEn
                 {
                     workerFunc = newWorkerFunc,
                     preservesPartitioning = preservesPartitioning && preservesPartitioningParam,
-                    previousRddProxy = this.previousRddProxy,
-                    prevSerializedMode = this.prevSerializedMode,
-
-                    sparkContext = this.sparkContext,
```
danielli90: "this." prefix seems to me helps readability - we may want to keep "this." prefix in this specific case.

PR #222, opened 2016-01-09
Simplify the `Split` expression

rdavisau:
Example looks cleaner if we don't use the `(string[], StringSplitOptions)` overload for a single character split :bowtie:

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,7 +6,7 @@ For example, the word count sample in Apache Spark can be implemented in C# as f
 
 ```c#
 var lines = sparkContext.TextFile(@"hdfs://path/to/input.txt");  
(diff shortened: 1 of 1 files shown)


skaarthik:
Thanks for the contribution @rdavisau . The sample code (https://github.com/Microsoft/SparkCLR/blob/master/csharp/Samples/Microsoft.Spark.CSharp/RDDSamples.cs#L514) also uses the same Split method. If you want to fix that as well, please send a PR.

skaarthik:
Just noticed that there are few other places in that .cs file where the same Split() overload is being used.

rdavisau:
Noted, I'll take a look :+1:

PR #223, opened 2016-01-09
Modify samples to use simpler `String.Split` overload where possible

rdavisau:
Applies change from #222 to the samples


rdavisau:
:skull: 
let me revisit the last invocation

PR #225, opened 2016-01-10
Use spark broadcast variables in the right way in SparkContextBroadcastSample

tawan0109:
In sample `SparkContextSamples.SparkContextBroadcastSample()`, broadcast variable `b` is used in following way:
```c#
 SparkContext.Parallelize(new[] { 0, 0 }, 1).FlatMap(new BroadcastHelper<int>(b.Value).Execute).Collect();
```
However, the right way to use Broadcast variable `b` should be like this:
```c#
SparkContext.Parallelize(new[] { 0, 0 }, 1).FlatMap(new BroadcastHelper<int>(b).Execute).Collect(); 
```


danielli90:
LGTM

PR #228, opened 2016-01-11
adding JDBC example to DataFrame API

skaarthik:
There was a post in StackOverflow asking for using JDBC API in SparkCLR. http://stackoverflow.com/questions/34494502/accessing-sql-server-data-into-sparkclr/34688154#34688154. Using the code from the post to create a new example.


danielli90:
LGTM

PR #240, opened 2016-01-13
release a sparkclr zip to github

danielli90:
This PR creates a sparkclr distribution and deploy to Github. The draft wiki page is here <https://github.com/danielli90/SparkCLR/wiki/Quick-Start>. A test release zip is here <https://github.com/danielli90/SparkCLR/releases/tag/v1.5.100-testrelease-25>

Runsamples.cmd and downloadtools.ps1 are moved into a `localscript` dir, so it can be packaged along with `run` directory and `example` directory.

Working on one more change to remove the step in changing sparkclr nuget versions for Pi. But submit this PR to have early review feedback.


appveyor.yml
```diff
@@ -41,21 +41,22 @@ after_test:
     - "SET PATH=C:\\Python34;C:\\Python34\\Scripts;%PATH%"
     - pip install codecov
     - codecov -f "SparkCLRCodeCoverage.xml"
-    - cmd: cd .\build
+    - cmd: cd .\build\localscript
```
skaarthik: an alternate name to consider is "localmode" - to make it self-explanatory

appveyor.yml
```diff
@@ -41,21 +41,22 @@ after_test:
...
   - path: '**\Microsoft*.nupkg'                 # find all NuGet packages recursively
   - path: 'scala\target\spark-clr*.jar'
+  - path: 'build\target\spark-clr*.zip'
 
 deploy:
   - provider: GitHub                            # deploy to GitHub releases
     auth_token:                                 
       secure: 4JIzL2DO3OXXjdrso+VhA0RjMvUAg8BnZRHXojswaryXFe9mDJ3ExjssASWL1FZq 
-    artifact: /Microsoft.*\.nupkg/,/spark-clr.*\.jar/
+    artifact: /spark-clr.*\.zip/
```
skaarthik: If we use a md file instead of wiki for the instructions, it could be included as a part of zip file

build/Build.cmd
```diff
@@ -128,12 +132,38 @@ copy /y Samples\Microsoft.Spark.CSharp\data\* "%SPARKCLR_HOME%\data\"
...
+    goto :distdone
+)
+
+set TARGET_FILE=.\run\scripts\sparkclr-submit.cmd
+@rem update sparkclr version in sparkclr-submit.cmd file
+powershell -NoProfile -ExecutionPolicy Bypass -Command "((Get-Content %TARGET_FILE%) -replace '\(set SPARKCLR_JAR=.*\)', '(set SPARKCLR_JAR=%SPARKCLR_NAME%.jar)') | Set-Content %TARGET_FILE% -force"
+
+@rem
+@rem Update SparkCLR package version to this release. Example in *.csproj:  
+@rem     <HintPath>..\packages\Microsoft.SparkCLR.1.5.2-SNAPSHOT\lib\net45\CSharpWorker.exe</HintPath>
```
skaarthik: I am not sure if this section is needed. We need to update the packages.config file in the repo just before creating a new release tag, With that out-of-the-box experience in using the examples will be smooth and running Build.cmd will not be prereq for successfully compiling examples.

build/Build.cmd
```diff
@@ -128,12 +132,38 @@ copy /y Samples\Microsoft.Spark.CSharp\data\* "%SPARKCLR_HOME%\data\"
...
+@rem
+set TARGET_FILE=..\examples\pi\pi.csproj
+powershell -NoProfile -ExecutionPolicy Bypass -Command "((Get-Content %TARGET_FILE%) -replace '\\Microsoft\.SparkCLR.*\\lib', '\Microsoft.SparkCLR.%ProjectVersion%\lib') | Set-Content -Encoding UTF8 -Path %TARGET_FILE% -force"
+
+@rem
+@rem Update SparkCLR package version to this release. Example in packages.config:  
+@rem     <package id="Microsoft.SparkCLR" version="1.5.2-SNAPSHOT" targetFramework="net45" />
+@rem
+set TARGET_FILE=..\examples\pi\packages.config
```
skaarthik: We will be adding more samples in addition to the two we already have. Basically anything that involves and external system (like DB, Kafka, Parquet, ...) can be included in examples.sln. So we do not want to update each project at the time of build. It can still be done using a script but at development time (not build).

appveyor.yml
danielli90: agreed.

appveyor.yml
danielli90: Initially, Wiki page is lot easier to adjust and make constant changes. Once Wiki page stabilized, agreed, we should convert wiki page content into a readme.md or some sort.

build/Build.cmd
danielli90: I hope this section can keep creating a new release tag as minimal as just tag creation, and defer the changes needed in packages.config when creating the distribution.

skaarthik:
Comments on Wiki:

* "... using SparkCLR M1 on Windows..." - M1 is a release artifact in the repo - let's try to avoid using that in the instructions and stick to versions
* "SparkCLR Maven Central package" - wouldn't it be better to just name the jar file?
* "SparkCLR Nuget package binaries and its dependencies" - same comment as above
* "PI_DIR=%SPARKCLR_HOME%\PI" - I am not sure what this is for. Do we need something similiar for all examples we will be adding?
* "Replace call sparkclr-submit.cmd... with call spark-submit.cmd --exe Pi.exe " - what is this for?
* Instead of re-purposing runsamples.cmd, we may want to include custom scripts for each sample to make the process/artifacts easy to understand. That will also give an opporutnity for us to perform custom steps - like including jdbc jar in classpath etc.

build/Build.cmd
skaarthik: we are changing files that are in the repo behind the scenes at the time of build. That might be very confusing.

build/Build.cmd
danielli90: Let's see. Update all packages.config and all *.proj is quite straightforward either in batch or in powershell script at build time. I expect examples will be mainly experimented by app developers who download SparkCLR distribution; and platform developers who work on SparkCLR core bits are expected to mainly spend time with samples but not examples. If this expectation turns out to be true, changing version stamps at build time can save some good amount of maintenance overhead for these examples.

appveyor.yml
danielli90: fixed

build/Build.cmd
danielli90: fixed by making a copy of examples, before changing versions. Also added a setversion.ps1 under "localmode", so platform developers can also change versions at development time.

build/Build.cmd
danielli90: Created a new setversion.ps1 under localmode, so platform devs can update versions at dev time as well. Especially when we fork a new branch, this script should help.

danielli90:
Regarding Wiki page - 

Synced up with @tawan0109. Will come up with a solution for running Pi example without asking developer to manually changing runsamples.cmd.

PR #244, opened 2016-01-13
rename "run" directory to "runtime"; runsamples.cmd to support examples

danielli90:
Prepare for M1 release -   
1. Rename `run` to `runtime` - will update wiki page suggesting to use `runtime` to submit SparkCLR jobs to Spark clusters (client and cluster modes);
2. @tawan0109 updated runsamples.cmd to support running examples locally; will update wiki page accordingly;
3. sparkclr-submit.cmd is updated to automatically set SPARKCLR_HOME, if possible; this will save one more step for developers who deploy "runtime" to the cluster.
4. `scripts\spark.conf` is moved to `build\localmode\spark.conf`;
5. `build\localmode\setversion.ps1` is replaced by `examples\SetSparkClrPackageVersion.ps1`, for better discovery and help developers to adjust the versions;
6. `examples\jdbcdataframe` has updated packages.config and .csproj file, with references to sparkclr  nuget package. In the final sparkclr distribution, an app dev is able to build both Pi and jdbcdataframe (and future examples) in Visual Studio directly.

Next, on to update Wiki page now.


build/localmode/spark.conf/spark-1.5.2-log4j.properties.template
```diff
@@ -0,0 +1,18 @@
+# Set everything to be logged to the console
```
skaarthik: With the addition of this file, should we delete 1.4.1 template above?

examples/SetSparkClrPackageVersion.ps1
```diff
@@ -0,0 +1,102 @@
+#
```
skaarthik: As discussion we can keep the version identifiers updated prior to release tagging

build/localmode/RunSamples.cmd
```diff
@@ -13,6 +19,13 @@ if "%1" == "" (
         set VERBOSE="verbose"
         @echo [RunSamples.cmd] VERBOSE is !VERBOSE!
     )
+
```
skaarthik: RunSamples already do too many things. I am not sure if we want to add more to it. For running examples instructions could just involve 3 steps - 
* the first step is optional to run precheck.cmd to download & set env. variable for spark, hadoop etc. if users do not have them installed
* the second step is for folks who skip step #1 - to set env. variable for spark, hadoop
* third step is to run a "sparkclr-submit ...." command

After for M1 or after, we can make these steps easy to follow by having a script called RunExample.cmd

build/localmode/spark.conf/spark-1.5.2-log4j.properties.template
danielli90: agreed.

examples/SetSparkClrPackageVersion.ps1
danielli90: agreed.

build/localmode/RunSamples.cmd
danielli90: One option is to keep the change in RunSamples for development benefits, but not offering RunSamples in the wiki page for running examples. So in the future, if we offer another script for running example, it will not introduce confusions,

skaarthik: Another option. Have RunSparkCLR.cmd that takes exe as the parameter. RunSamples can just call RunSparkCLR with SparkCLRSamples.exe as arg. For other scenario is including examples, users may use RunSparkCLR directly.

skaarthik: Both are reasonable options IMO.

danielli90:
TravisCi passed, without any changes related to TravisCI. Please review this PR now.

xiongrenyi:
LGTM

PR #246, opened 2016-01-14
Improve test coverage in 'utils' & remove unused methods in SerDe.cs

tawan0109:
* Improve test coverage of `SerDe.cs`
* Improve test coverage of `PayloadHelper.cs`
* Remove unused method `SerDe.Convert(int value)`, `SerDe.Convert(long value)` and `SerDe.Convert(double value)`

### DIFF
--- a/csharp/AdapterTest/PayloadHelperTest.cs
+++ b/csharp/AdapterTest/PayloadHelperTest.cs
@@ -1,0 +1,327 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using System.Collections.Generic;
+using System.IO;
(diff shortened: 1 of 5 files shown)


csharp/AdapterTest/PayloadHelperTest.cs
```diff
@@ -0,0 +1,327 @@
...
+        [Test]
+        public void Test()
+        {
+            const string expectedClassName = "class1";
+            const string expectedMethod = "method1";
+            object[] expectedParams =
+            {
+                1, // int 
+                2L, // long
+                3D, // double
```
jthelin: You might want to add some negative values in here too, for completeness?

tawan0109: Thanks for the comment, will send a PR for it.

PR #257, opened 2016-01-14
RunSamples.cmd should signal failure if non-zero RC from sparkclr-submit script

jthelin:
This tweak to `RunSamples.cmd` script makes it easier to detect any problems in the build environment, because build agent [eg VSO] will detect non-zero exit code which was previously being masked by extra commands at end of script.

- `RunSamples.cmd` should signal runtime failure if non-zero RC from underlying call to `sparkclr-submit.cmd` script. [The final `popd` was previously masking this value.]

- Fix non-standard usage of quotes between RunSamples.cmd and PreCheck.cmd scripts.

- Print directory location info for strategic pushd's

Should have no affect on the core team's Appveyor build, but my private VSO build was not detecting / marking build failures when it should have been due to some test config issues.

PR #270, opened 2016-01-20
M2 build settings and mvn --quiet

jthelin:
I have found these changes were required to get SparkCLR building again on my VSO instance, mostly due to problems with mvn settings. 

PR #257 commit 7159d9191f231e6f9e4b527322152d15fa1e2594 helped unmask the VSO build problems, which had been happening for a while.

The main part of the change is this:

- If mvn not already installed, then set M2 env variables from download 
d4310ce4bd243a77872c989b14ba195720aba4d1

This change set also contains a tweak to run mvn in --quiet more on the build servers, because the massive amount of download progress messages in the build log were also causing any build issues to be hard to spot.

- Run Maven in `--quiet` mode to avoid all the download progress messages in the build server logs.
c3eaa3afe31371a7d1fc5144f1e5d184c4116368

The other parts of this change set are minor diagnosability tweaks and/or clean-ups.

I have left the commits as separate micro-changes so that it is easier to see what each step does, but let me know if you prefer squash before merge.

### DIFF
--- a/build/Build.cmd
+++ b/build/Build.cmd
@@ -88,6 +88,8 @@ if %ERRORLEVEL% NEQ 0 (
 
 @echo SparkCLR Scala binaries
 copy /y target\*.jar "%SPARKCLR_HOME%\lib\"
+@REM Exclude any original pre-shaded .jar files
+del /q "%SPARKCLR_HOME%\lib\original-*"
 popd
 
 @REM Any .jar files under the lib directory will be copied to the staged runtime lib tree.
--- a/build/Build.cmd
+++ b/build/Build.cmd
@@ -37,7 +37,7 @@ if NOT EXIST "%SPARKCLR_HOME%\samples" mkdir "%SPARKCLR_HOME%\samples"
 pushd "%CMDHOME%\..\scala"
 
 @rem clean the target directory first
-call mvn.cmd clean
+call mvn.cmd %MVN_QUIET% clean
 
 @rem
 @rem Note: Shade-plugin helps creates an uber-package to simplify SparkCLR job submission;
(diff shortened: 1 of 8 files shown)


build/Build.cmd
```diff
@@ -81,13 +81,15 @@ IF "%APPVEYOR_REPO_TAG%" == "true" (goto :sign)
...
-    popd
-    goto :eof
+  @echo Build SparkCLR Scala components failed, stop building.
+  popd
+  goto :eof
 )
 
 @echo SparkCLR Scala binaries
 copy /y target\*.jar "%SPARKCLR_HOME%\lib\"
+@REM Exclude any original pre-shaded .jar files
```
skaarthik: What is this change for?

build/localmode/RunSamples.cmd
```diff
@@ -8,11 +8,6 @@ set CMDHOME=%CMDHOME:~0,-1%
 set VERBOSE=
 set USER_EXE=
 
-set CMDHOME=%~dp0
-
-@REM Remove trailing backslash \
-set CMDHOME=%CMDHOME:~0,-1%
```
skaarthik: Is this block not needed if we use quotes?

jthelin: The code is already in this file at line 3, so this is removing the duplicate code.

build/Build.cmd
jthelin: When i run the build\Build.cmd script, there are two .jar files placed in the runtime\lib directory - the "original" [pre-shaded] and the final [uber / shaded] jar.

I believe we only want the latter in the runtime directory tree?

```
Directory of E:\Depot\GitHub\SparkCLR\build\runtime\lib

01/21/2016  09:50 AM           159,661 original-spark-clr_2.10-1.6.0-SNAPSHOT.jar
01/21/2016  09:50 AM        32,835,116 spark-clr_2.10-1.6.0-SNAPSHOT.jar
```

danielli90: yes - we only want the shaded version under "runtime". How about changing - 
copy /y target\*.jar "%SPARKCLR_HOME%\lib\" 
To - 
copy /y target\spark*.jar "%SPARKCLR_HOME%\lib\"

jthelin: OK, I'll push update to this PR for that way.

danielli90:
LGTM

PR #272, opened 2016-01-21
fix conf argument type in the proxy to Hadoop API methods

michael-stucco:
The prototype for all the Hadoop API mehtods is expecting a `java.util.HashMap` as the `conf` argument, however now a `java.util.Hashtable` is passed and the invocation fails with:

```java
cannot find matching method class org.apache.spark.api.python.PythonRDD.newAPIHadoopRDD. Candidates are:
newAPIHadoopRDD(class org.apache.spark.api.java.JavaSparkContext,class java.lang.String,class java.lang.String,class java.lang.String,class java.lang.String,class java.lang.String,class java.util.HashMap,int)
newAPIHadoopRDD on object of type NullObject failed
No matched method found for class org.apache.spark.api.python.PythonRDD.newAPIHadoopRDD
java.lang.Exception: No matched method found for class org.apache.spark.api.python.PythonRDD.newAPIHadoopRDD
        at org.apache.spark.api.csharp.CSharpBackendHandler.handleMethodCall(CSharpBackendHandler.scala:150)
...
```

This PR is fixing it,
Thanks!

### DIFF
--- a/csharp/Adapter/Microsoft.Spark.CSharp/Proxy/Ipc/SparkContextIpcProxy.cs
+++ b/csharp/Adapter/Microsoft.Spark.CSharp/Proxy/Ipc/SparkContextIpcProxy.cs
@@ -152,7 +152,7 @@ namespace Microsoft.Spark.CSharp.Proxy.Ipc
 
         public IRDDProxy NewAPIHadoopFile(string filePath, string inputFormatClass, string keyClass, string valueClass, string keyConverterClass, string valueConverterClass, IEnumerable<KeyValuePair<string, string>> conf, int batchSize)
         {
-            var jconf = GetJavaMap<string, string>(conf);
+            var jconf = GetJavaHashMap<string, string>(conf);
             var jvmRddReference = new JvmObjectReference((string)SparkCLRIpcProxy.JvmBridge.CallStaticJavaMethod("org.apache.spark.api.python.PythonRDD", "newAPIHadoopFile",
                 new object[] { jvmJavaContextReference, filePath, inputFormatClass, keyClass, valueClass, keyConverterClass, valueConverterClass, jconf, batchSize }));
             return new RDDIpcProxy(jvmRddReference);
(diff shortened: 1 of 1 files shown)


csharp/Adapter/Microsoft.Spark.CSharp/Proxy/Ipc/SparkContextIpcProxy.cs
```diff
@@ -361,6 +361,18 @@ public IColumnProxy CreateWindowFunction(string name)
             }
             return jmap;
         }
+
+        public static JvmObjectReference GetJavaHashMap<K, V>(IEnumerable<KeyValuePair<K, V>> enumerable)
```
skaarthik: Would it be better to use a dictionary instead of IEnumerable<KVP>? With that, we do not have to check for duplicate keys.

skaarthik:
Thanks for your contribution @michael-stucco. I have just added a minor comment/question on your change.

csharp/Adapter/Microsoft.Spark.CSharp/Proxy/Ipc/SparkContextIpcProxy.cs
michael-stucco: I agree that a `Dictionary` (or actually `Dictionary<sting,string>`) seems exactly what is needed for the `conf ` argument. However the same is true for the previously used `GetJavaMap` (which created a `java.util.Hashtable`), and maybe this optimization you suggest should be done in an additional/different PR?
Thanks!

danielli90:
LGTM. Agreed to fix the failure first and optimize later.

csharp/Adapter/Microsoft.Spark.CSharp/Proxy/Ipc/SparkContextIpcProxy.cs
skaarthik: I didn't realize that the signature change would need change to interfaces.

PR #273, opened 2016-01-21
Fix Pi example can't run in yarn-cluster mode issue

tawan0109:
Checkpoint directory needs to be a HDFS path when the deploy mode is `yarn-cluster`. Remove `SparkContext.SetCheckpointDir(Path.GetTempPath());` to make `Pi` example run well in `yarn-cluster` mode.


skaarthik:
LGTM

PR #287, opened 2016-01-28
fix Kafka createDirectStream API name and add MapWithState API for CSharpDStream

xiongrenyi:


danielli90:
Talked to @xiongrenyi. It will be great to add reference to Scala API. Otherwise, LGTM.

PR #288, opened 2016-02-01
spark-xml example and updated SparkCLR version references

skaarthik:


examples/pi/Pi.csproj
```diff
@@ -41,7 +41,7 @@
     </Reference>
     <Reference Include="CSharpWorker">
       <SpecificVersion>False</SpecificVersion>
-      <HintPath>..\packages\Microsoft.SparkCLR.1.6.0-SNAPSHOT\lib\net45\CSharpWorker.exe</HintPath>
+      <HintPath>..\packages\Microsoft.SparkCLR.1.5.20\lib\net45\CSharpWorker.exe</HintPath>
```
danielli90: typo. should be "1.5.200"

examples/pi/Pi.csproj
```diff
@@ -8,7 +8,7 @@
     <OutputType>Exe</OutputType>
     <AppDesignerFolder>Properties</AppDesignerFolder>
     <RootNamespace>Microsoft.Spark.CSharp.Examples</RootNamespace>
-    <AssemblyName>Pi</AssemblyName>
+    <AssemblyName>SparkCLRPiExample</AssemblyName>
```
danielli90: The name of "SparkCLRPiExample" is a bit too long. Scala example is called "SparkPi", python example is called "Pi.py". If we prefer to stick with Scala notation, we can use SparkPi then (similar to sparkxml above)?

examples/pi/App.config
```diff
@@ -34,6 +34,19 @@
...
+  <appSettings>
+    <!--********************************************************************************************************-->
+    <!--** Uncomment the following settings to run Spark driver executable in **local** or **debug** modes ** -->
+    <!--** In debug mode, the driver is not launched by CSharpRunner but launched from VS or command prompt not configured for SparkCLR ** -->
+    <!--** CSharpBackend should be launched in debug mode as well and the port number from that should be used below ** -->
+    <!--** Command to launch CSharpBackend in debug mode is "sparkclr-submit.cmd debug" ** -->
+    <!--********************************************************************************************************-->
+
+    <!--
+    <add key="CSharpWorkerPath" value="C:\Git\SparkCLR\examples\Pi\bin\Debug\CSharpWorker.exe"/>
```
danielli90: In the case of example, exe is under "..\packages\Microsoft.SparkCLR.1.5.20\lib\net45\CSharpWorker.exe". It will be ideal to change the sample line pointing to this one under "..\packags".

examples/pi/Pi.csproj
skaarthik: will fix it

examples/pi/Pi.csproj
skaarthik: Will use SparkClrPi as the name

examples/pi/App.config
skaarthik: relative path will not work

skaarthik: ...and the path to worker exe should also contain the exe that is submitted to SparkCLR.

danielli90: Ignore my previous comment - CSharWorker.exe is actually copied over to <Pi\bin\debug|release> when Pi example is built/compiled.

PR #293, opened 2016-02-03
documentation fix: built binaries are under build/runtime/ instead of run/

cyruszhang:


skaarthik:
LGTM. Thanks @cyruszhang

Issue #304, opened 2016-02-16
WorkerTest unit tests fail when using R# plugin with Visual Studio


cyruszhang:
If using NUnit3 Adapter, then the unit tests will succeed.

The problem is that location specified by WorkerTest->CreateServer->Assembly.GetExecutingAssembly() does not have CSharpWorker.exe. Resharper initiates NUnit tests in a different dir than the NUnit Adapter.

Test initialization code should be dev tool agnostic and succeed whether we are using R# or not.

cyruszhang:
![image](https://cloud.githubusercontent.com/assets/2977989/13086468/36662c2e-d498-11e5-80f1-1ccbb7be6edd.png)

cyruszhang:
Assembly.GetExecutingAssembly() when using R#:
![image](https://cloud.githubusercontent.com/assets/2977989/13086519/67606ee8-d498-11e5-8e10-fa3e09f63a05.png)

jthelin:
Try turning off the "Shadow Copy" option in ReSharper unit test options.

ReSharper -> Options -> Tools -> Unit Testing

[ ] Shadow-copy assemblies being tested.

cyruszhang:
Yes, that works. A simple code change would make this problem go away as
well.

On Tue, Feb 16, 2016 at 1:02 PM, Jorgen Thelin <notifications@github.com>
wrote:

> Try turning off the "Shadow Copy" option in ReSharper unit test options.
>
> ReSharper -> Options -> Tools -> Unit Testing
>
> [ ] Shadow-copy assemblies being tested.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/Microsoft/SparkCLR/issues/304#issuecomment-184873553>.
>



-- 
Cyrus, Yilei Zhang
+1 (312) 813 6818

skaarthik:
Thanks for fixing this issue, @cyruszhang

PR #305, opened 2016-02-16
Fix R# Unit Tests error when shadow copying is enabled

cyruszhang:
image


cyruszhang:
Reference: 
http://www.nunit.org/index.php?p=settingsDialog&r=2.6.3
Shadow Copy

cyruszhang:
https://github.com/Microsoft/SparkCLR/issues/304

PR #308, opened 2016-02-19
Improve RDDCheckpointSample and add validations

tawan0109:
In current `RDDCheckpointSample`, `Console.WriteLine(rdd.IsCheckpointed)` always print out `false`.
Call `Count()` action of the RDD to make sure `doCheckpoint` method invoked and `rdd.IsCheckpointed` returns true.


skaarthik:
LGTM

PR #309, opened 2016-02-23
Add '--checkpointdir' to explicity specify checkpoint directory when run samples

tawan0109:
* Currently in SparkCLR samples checkpoint directory is set to a local temp path.
* Samples breaks when it is running in yarn mode.
* Add new parameter `--checkpointdir` to explicitly specify a checkpoint directory(HDFS path for yarn mode).
* To be backward compatible, checkpoint directory is set to a local temp path when `--checkpointdir` is not specified.

### DIFF
--- a/csharp/Samples/Microsoft.Spark.CSharp/CommandlineArgumentProcessor.cs
+++ b/csharp/Samples/Microsoft.Spark.CSharp/CommandlineArgumentProcessor.cs
@@ -59,6 +59,11 @@ namespace Microsoft.Spark.CSharp.Samples
                 {
                     configuration.IsValidationEnabled = true;
                 }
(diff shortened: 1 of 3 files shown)


danielli90:
LGTM. 

Down the road, we need to have an example to show checkpoint usage, similar to this one - https://github.com/apache/spark/blob/master/examples/src/main/scala/org/apache/spark/examples/streaming/RecoverableNetworkWordCount.scala

tawan0109:
Yes, we need this example. But I think streaming sample `DStreamTextFileSamples` has already showed checkpoint usage. This sample can be restored from checkpoint directory.

Of course, If needed, we might want to make some changes to `DStreamTextFileSamples` to make the checkpoint directory can be configured instead of using a temp directory, and add some console outputs to indicate whether current `StreamingContext` is loaded from checkpoint.

PR #331, opened 2016-03-02
enable push to myget.org with keeping pushing to nuget.org on tag commit

jayjaywg:


danielli90:
LTGM. Thanks @jayjaywg

PR #335, opened 2016-03-02
use myget feed mobiusforspark

jayjaywg:


xiongrenyi:
Hi Guan,

unlike our official release in github, current myget only get build of c# side and to my understanding scala side jar gets abandoned.

is it possible and/or does it make sense to include Mobius jar in myget as well?

Thanks.

PR #341, opened 2016-03-07
add new API DirectKafkaStreamWithRepartition manual test only in Sandbox

xiongrenyi:
customizing KafkaUtils using inheritance model with dynamic kafka rdd partitioning and kafka read fault tolerance. no automated tests yet


skaarthik:
We need unit test for C# code to make sure code cov % does not drop.

Can we also get unit tests for the scala code from Spark tree and reuse them for the custom implementation?

xiongrenyi:
unit test added initially for the new direct kafka API like the existing one, it's the proxy code (not covered in unit test) which drops the cov %

xiongrenyi:
want to unblock people, so only manual testing for now

xiongrenyi:
added sample (functional test) as a separately

PR #342, opened 2016-03-08
add sample (functional test) for the new API DirectKafkaStreamWithRepartition

xiongrenyi:


xiongrenyi:
Just realized adding validation to streaming sample which never stops needs some more work.
 
Since the sample has some implicit validation, considering the immediate priority is to have some validation instead of nothing, I prefer to merge the PR first and add some stronger validation tomorrow.
 
What do you think?

skaarthik:
If it can be done tomorrow, it is better to wait. If it is going to take much longer to implement better validation, then we need to look into our options and see which one would make sense.

xiongrenyi:
maybe one more day, but not much longer for sure

sharing context missing above:

Kaarthik and I had an offline discussion on dropped code coverage and functional test.
1. code coverage actually increased but the number showing low somehow.
2. for test, we first explored the idea of borrowing existing scala test, but find out not very straight forward and quite a bit effort needed, so we agreed to add a sample with validations like other tests

xiongrenyi:
adding validation means more comprehensive test by adding fixed data set to kafka test topic, fortunately, my local kafka instance created long time ago still working.

also made needed GetNumPartitions public which was made the same from Spark 1.6

last, fixed some minor bugs including from tqin's

skaarthik:
LGTM

xiongrenyi:
1. I have added detailed comments for async metadata refresh after sync with tqin, 
2. moved test settings to app.config in dstream kafka sample
3. reverted some changes to DynamicPartitionKafkaRDDIterator's error handling and added more logs

skaarthik:
LGTM

PR #344, opened 2016-03-14
refactored worker and fixed issue in perf suite

skaarthik:


qintao1976:
LGTM.
Please check in the change. Then I can send the codeflow of useDaemon C# worker process based on the new code base.
Thanks

PR #356, opened 2016-03-22
Make the name of 'referesh-offsets' thread unique for different DynamicPartitionKafkaInputDStream instances

tawan0109:
* Currently the thread names for all`refreshOffsetsScheduler` instances are same, which can't be distinguished in logs.
* Set the name to `refresh-offsets-%d` won't solve the problem.
* For each thread name, append a random uuid (only use first 8 chars).


qintao1976:
LGTM

tawan0109:
The coverage decrease should be a false alarm, as this PR did not touch C# codes.

PR #358, opened 2016-03-24
improving assert to get details when test fails

skaarthik:


skaarthik:
This PR helped get the worker output and identify what's going on but no longer needed. @qintao1976 fixed the issue in https://github.com/Microsoft/Mobius/commit/6ae3a81196f6a4e1c63556d092bba9844653a7c3

PR #359, opened 2016-03-24
In MultiThreadWorker mode, does not let the whole worker process exit when some TaskRunner fails

qintao1976:
1. In MultiThreadWorker mode, does not let the whole worker process exit when some TaskRunner fails.
2. Minor change to improve log info in worker


danielli90:
LGTM

PR #361, opened 2016-03-25
Fix the UT - TestBroadcastVariablesInWorker

qintao1976:
The test of assertMessage should be postponed after worker exit.


skaarthik:
LGTM

PR #363, opened 2016-03-28
Fix the problem that TaskRunner's log4j logger is not properly initia…

qintao1976:
…lized

Because log4j can't be initialized early (to avoid flush stdout because JVM will read port info via pipe), the logger can't be initialized statically.


tawan0109:
LGTM

PR #364, opened 2016-03-29
Add the option of use offset size to control dynamic kafka partition

qintao1976:


scala/src/main/org/apache/spark/streaming/api/kafka/DynamicPartitionKafkaRDD.scala
```diff
@@ -48,7 +48,22 @@ class DynamicPartitionKafkaRDD[
   ) extends KafkaRDD[K, V, U, T, R](sc, kafkaParams, offsetRanges, leaders, messageHandler) {
   override def getPartitions: Array[Partition] = {
     if (numPartitions == 0) {
-      super.getPartitions
+      val steps = offsetRanges.groupBy(o => o.topic).map { case (topic, _) =>
```
skaarthik: Unit tests?

danielli90:
Need to expand the comment for "numPartition" to explain "spark.streaming.kafka.maxRatePerTask.<Topic>", when numPartition is zero. Other than that. LGTM.

qintao1976:
Add comments and add UT in scala side.

jthelin:
@qintao1976 FYI, you don't need to delete + recreate the PR branch in your repository to add more changes to a Pull Request. 

If you just push additional commits to your existing PR branch then GitHub includes them incrementally.

You can also do a rebase/squash and then _force push_ to the _same_ PR branch if you want to clean up the commit history for a PR for any reason -- such as when some feedback / discussion during a PR might result in additional changes which would make parts of the original code confusing to anyone looking at the change later.

HTH.

xiongrenyi:
numPartitions should have default value -1

CreateDirectStreamWithRepartition(StreamingContext ssc, List<string> topics, Dictionary<string, string> kafkaParams, Dictionary<string, long> fromOffsets, int numPartitions = -1)

And getPartitions:

If numPartitions = -1, either repartition based on spark.streaming.kafka.maxRatePerTask or do nothing if config not defined
If numPartitions = 0, repartition using original kafka partition count
If numPartitions > 0, repartition using this parameter

  val steps = offsetRanges.groupBy(o => o.topic).map { case (topic, offRanges) =>
    if (numPartitions < 0)
      (topic, sparkContext.getConf.getInt("spark.streaming.kafka.maxRatePerTask." + topic, Int.MaxValue))
    else {
      val partitions = if (numPartitions == 0) offRanges.length else numPartitions
      // step can't be zero, use Math.max(..., 1) to ensure that step >= 1
      (topic, Math.max(offRanges.map(o => o.untilOffset - o.fromOffset).sum / partitions, 1))
    }
  }
  offsetRanges.flatMap { case o =>
    (o.fromOffset until o.untilOffset by steps(o.topic)).map(s =>
      (o.topic, o.partition, s, math.min(o.untilOffset, s + steps(o.topic)))
    )
  }.zipWithIndex.map { case ((topic, partition, start, until), i) =>
    val (host, port) = leaders(TopicAndPartition(topic, partition))
    new KafkaRDDPartition(i, topic, partition, start, until, host, port)
  }.toArray
}

qintao1976:
@jthelin , thanks for the advice!
@xiongrenyi , have revised according to your suggestion.

Thanks

PR #365, opened 2016-03-29
Add Init script to set up development shell environment

hebinhuang:
Adding Init script to set up the proper PATH and ENV to use Github shell and build environment.
1. After running Init.cmd, it will create a powershell shortcut for Mobius project in which you can use posh-git shell and build directly from that window. It also to call \build\localmode\downloadtools.ps1 to download build tool.
2. When you click the shortcut for Mobius from your desktop, it will do the following jobs:
    - Import GitHub Shell ENV
    - Import EVN of Developer Command Prompt for VS
    - Create batch of alias from build\aliases (e.g. n for notepad, .. for cd .., gco for git checkout)

### DIFF
--- a/Init.cmd
+++ b/Init.cmd
@@ -1,0 +1,8 @@
+@echo off
+@rem Windows 7/8/10 may not allow powershell scripts by default
+powershell -Command Set-ExecutionPolicy -Scope CurrentUser Unrestricted
+
+@rem Setup init environment
+pushd "%CMDHOME%"
+powershell -f Init.ps1 -SetupEnv
+popd
--- a/Init.ps1
+++ b/Init.ps1
@@ -35,7 +35,7 @@ process
         }
     }
 
-    $Root = Get-ScriptDirectory
+    $Global:Root = Get-ScriptDirectory
 
     function Create-ShortCut {
         try
--- a/Init.ps1
+++ b/Init.ps1
@@ -1,0 +1,126 @@
+# Copyright (c) Microsoft. All rights reserved.
+# Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+<#
+.SYNOPSIS
+    Sets up and initialize the developer Shell Environment for Mobius
+.DESCRIPTION
+    Sets up the proper PATH and ENV to use Posh-Git shell and build environment
+#>
+[CmdletBinding()]
+Param(
+    [Parameter(Mandatory = $false)]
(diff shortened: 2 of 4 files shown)


build/aliases/aliases.txt
```diff
@@ -0,0 +1,28 @@
+root	= pushd $Root\
+build	= $Root\build\build.cmd
+
+..	= cd ..
+n	= notepad.exe
+n++	= notepad++.exe
+
+ga	= git add
```
skaarthik: I am not sure if we need to make it available in the repo. Not everyone will use aliases. It boils down to individual preferences. Since it is not core to developing Mobius, I recommend not adding it to repo

Init.cmd
```diff
@@ -0,0 +1,8 @@
+@echo off
```
skaarthik: Instead of creating a new folder for the new files, consider reusing dev folder

Init.ps1
```diff
@@ -0,0 +1,126 @@
...
+        write-debug "Defined function:`n$func"
+        $func | iex
+        $newAlias = Set-Alias -Name $alias -Value "Alias$alias" -Description "$cmd" -Option AllScope -scope Global -passThru
+    }
+
+    if ($SetupEnv) {
+        # Check if the GitHub for Windows is installed
+        if (!(Test-Path "$env:LOCALAPPDATA\GitHub"))
+        {
+            $Host.UI.WriteErrorLine("The development enviroment requires ""GitHub Desktop"" be installed. You can download from https://desktop.github.com/")
```
skaarthik: GitHub Desktop is not mandatory for Mobius development.

Init.ps1
```diff
@@ -0,0 +1,126 @@
...
+        }
+        else
+        {
+            $Invocation.InvocationName.Substring(0,$Invocation.InvocationName.LastIndexOf("\"));
+        }
+    }
+
+    $Root = Get-ScriptDirectory
+
+    function Create-ShortCut {
```
skaarthik: Creating shortcut is also unnecessary IMO and not core to Mobius development. It might be simpler just to stick to launching the dev prompt. Users can create a shortcut to Init script if needed.

hebinhuang:
I agree with you that this is totally not related with our project. It is just a individual preferences. I will put it into my personal project and wiki it if some one want use it to set up a shell environment.

PR #367, opened 2016-03-30
Fix build warnings (part 2): Fix special characters in XML comments

hebinhuang:
Fix special characters in XML comments and add XML coments for PairRDDFunctions, RDD, SparkContext, and SqlContext


xiongrenyi:
LGTM

PR #368, opened 2016-03-31
https://github.com/Microsoft/Mobius/issues/267

xsidurd:
Need a wordcount example similar to https://github.com/apache/spark/blob/master/examples/src/main/python/wordcount.py


xiongrenyi:
LGTM

PR #376, opened 2016-04-08
moving examples projects under subfolders (Batch, Sql and Streaming)

skaarthik:


skaarthik:
Just moved example projects to subfolders. No functional/code change and no impact to codecoverage. The failure in CodeCov check is not correct. Proceeding with the merge.

PR #377, opened 2016-04-08
Hivecontext

jayjaywg:


skaarthik:
@jayjaywg - note that the other samples that support HDFS [like sparkContext.TextFile()] rely on Spark API that also supports local path. That may not be true with the API that throws up in the Hive sample. Configuring and getting HDFS/Hive services working in AppVeyor and Travis may not be trivial. If you cannot get it to work with minimal effort, I suggest moving your sample to an example project (like what have for EventHub or JDBC) with [instructions](https://github.com/Microsoft/Mobius/blob/master/notes/running-mobius-app.md#running-examples-in-local-mode).

jayjaywg:
@skaarthik Thanks for the suggestion, I'm trying to use 'chmod' to modify the permissions of the directory. If that's not working, I'll consider moving to Example project.

PR #381, opened 2016-04-12
Skip existing files when unzip driver zip file to destination directory

tawan0109:


danielli90:
LGTM

tawan0109:
As only Scala codes are changed in this PR, don't think the test coverage drop is related to current PR.

Issue #274, opened 2016-01-22
improve unit test coverage in "Sql" namespace


skaarthik:
Some classes still have code coverage less than 75%. We need to increase the coverage to 75% or more.

See https://codecov.io/github/Microsoft/Mobius/csharp/Adapter/Microsoft.Spark.CSharp/Sql

skaarthik:
@hebinhuang - looks like the following files still have lots of blocks missing CC. Are they hard to cover or not important?

- https://codecov.io/github/Microsoft/Mobius/csharp/Adapter/Microsoft.Spark.CSharp/Sql/DataFrameNaFunctions.cs?ref=702d6d624f436da0a51a37315d455b38fbc234f8
- https://codecov.io/github/Microsoft/Mobius/csharp/Adapter/Microsoft.Spark.CSharp/Sql/Types.cs?ref=702d6d624f436da0a51a37315d455b38fbc234f8
- https://codecov.io/github/Microsoft/Mobius/csharp/Adapter/Microsoft.Spark.CSharp/Sql/Functions.cs?ref=702d6d624f436da0a51a37315d455b38fbc234f8

skaarthik:
and this as well...https://codecov.io/github/Microsoft/Mobius/csharp/Adapter/Microsoft.Spark.CSharp/Sql/DataFrame.cs?ref=702d6d624f436da0a51a37315d455b38fbc234f8

hebinhuang:
There is a part 2 to cover these files - see https://github.com/Microsoft/Mobius/pull/395
Columns.cs,DataFrame.cs,Types.cs,DataFrameNaFunctions.cs,Functions.cs

PR #384, opened 2016-04-16
#274: improve unit test coverage in Sql namespace

hebinhuang:
This is for M2 workitem #274 to add unit test to improve code coverage in "Sql" namespace

### DIFF
--- a/csharp/Adapter/documentation/Mobius_API_Documentation.md
+++ b/csharp/Adapter/documentation/Mobius_API_Documentation.md
@@ -142,6 +142,13 @@
             Used for collect operation on RDD
             
         
+###<font color="#68228B">Microsoft.Spark.CSharp.Core.IRDDCollector</font>
+####Summary
+  
+            
+            Interface for collect operation on RDD
+            
+        
 ###<font color="#68228B">Microsoft.Spark.CSharp.Core.DoubleRDDFunctions</font>
 ####Summary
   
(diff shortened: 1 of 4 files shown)


csharp/Adapter/documentation/Mobius_API_Documentation.md
```diff
@@ -142,6 +142,13 @@
             Used for collect operation on RDD
```
skaarthik: You could exclude changes in this file because you are not adding/removing any method from the public API

skaarthik: ...and also the XML file

csharp/AdapterTest/SerDeTest.cs
```diff
@@ -97,5 +95,155 @@ public void TestReadBytes()
...
+
+            return new RowImpl(new object[] { seq, "id " + seq, "name" + seq }, DataType.ParseDataTypeFromJson(jsonSchema) as StructType);
+        }
+    }
+
+    /// <summary>
+    /// Used to pickle StructType objects
+    /// Reference: StructTypePickler from https://github.com/apache/spark/blob/master/sql/core/src/main/scala/org/apache/spark/sql/execution/python.scala#L240
+    /// </summary>
+    internal class StructTypePickler : IObjectPickler
```
skaarthik: Can we reuse the picklers from WorkerTest?

csharp/Adapter/documentation/Mobius_API_Documentation.md
hebinhuang: Sure, I will move out these two files from the PR.

csharp/AdapterTest/SerDeTest.cs
hebinhuang: Can the WorkerTest be a dependency of AdapterTest? if so, we can reuse it.

PR #389, opened 2016-04-19
Improve unit test coverage in SQL namespace

hebinhuang:
1. Add unit tests to improve code coverage for "SQL" namespace
2. Create a new common library that shared by AdapterTest and WorkerTest. And move common code to the new library.


skaarthik:
LGTM

Issue #289, opened 2016-02-01
examples folder in release zip does not have binaries


skaarthik:
The release zip file includes source files under "examples" folder. Source files need to be replaced with binaries that can be run directly without the need to build.

https://github.com/Microsoft/SparkCLR/releases/download/v1.5.200/spark-clr_2.10-1.5.200.zip

skaarthik:
With https://github.com/Microsoft/SparkCLR/pull/294 example projects are built. Need to include them in "runtime" folder that will be used for release.

hebinhuang:
Checked in #400

skaarthik:
@hebinhuang - the latest release https://github.com/Microsoft/Mobius/releases/tag/v1.6.100-PREVIEW-1 does not seem to have example binaries. Can you take a look? You can refer to the logs of build 1.6.1-SNAPSHOT.685 in AppVeyor to troubleshoot.

hebinhuang:
It caused by the failure of compiling Examples.

HdfsWordCount -> C:\projects\sparkclr\examples\Streaming\HdfsWordCount\bin\Debug\SparkClrHdfsWordCount.exe
 1204C:\windows\Microsoft.NET\Framework\v4.0.30319\Microsoft.Common.targets(1605,5): warning MSB3245: Could not resolve this reference. Could not locate the assembly "CSharpWorker". Check to make sure the assembly exists on disk. If this reference is required by your code, you may get compilation errors. [C:\projects\sparkclr\examples\Sql\JdbcDataFrame\JdbcDataFrame.csproj]
 1205C:\windows\Microsoft.NET\Framework\v4.0.30319\Microsoft.Common.targets(1605,5): warning MSB3245: Could not resolve this reference. Could not locate the assembly "Microsoft.Spark.CSharp.Adapter". Check to make sure the assembly exists on disk. If this reference is required by your code, you may get compilation errors. [C:\projects\sparkclr\examples\Sql\JdbcDataFrame\JdbcDataFrame.csproj]
 1206Program.cs(6,30): error CS0234: The type or namespace name 'Core' does not exist in the namespace 'Microsoft.Spark.CSharp' (are you missing an assembly reference?) [C:\projects\sparkclr\examples\Sql\JdbcDataFrame\JdbcDataFrame.csproj]
[...]

PR #399, opened 2016-04-21
Fix Build.cmd to include examples binaries in release zip file

hebinhuang:
The release zip file includes source files under "examples" folder. the fix is to include examples binaries in release zip file.

Refer to #289


build/Build.cmd
```diff
@@ -142,16 +168,16 @@ xcopy /e /y "%CMDHOME%\..\scripts"  "%SPARKCLR_HOME%\scripts\"
 pushd "%CMDHOME%"
 if not exist ".\target" (mkdir .\target)
 
+set ProjectVersion=1.6.0-preview-3
```
skaarthik: why is this needed?

build/Build.cmd
```diff
@@ -142,16 +168,16 @@ xcopy /e /y "%CMDHOME%\..\scripts"  "%SPARKCLR_HOME%\scripts\"
...
+
 if not defined ProjectVersion (
     powershell -f .\localmode\zipdir.ps1 -dir "%SPARKCLR_HOME%" -target ".\target\runtime.zip"
     goto :distdone
 )
 
-set SPARKCLR_NAME=spark-clr_2.10-%ProjectVersion%
```
skaarthik: ...and this change?

build/Build.cmd
hebinhuang: It looks like the merge between my private branch and master has problems. this line is just for my test propose. Actually, I have removed the line before commit. Not sure why it came back.

Let me do a clean branch again.

PR #400, opened 2016-04-21
Fix Build.cmd to include examples binaries in release zip file

hebinhuang:
The release zip file includes source files under "examples" folder. the fix is to include examples binaries in release zip file.

Refer to #289


skaarthik:
LGTM

PR #404, opened 2016-04-21
Fix Build.cmd to stop building when Examples compiling failed

hebinhuang:
Currently, the build process cannot be stopped if Examples solution has compiling failure. The fix is to stop building process if Examples failed.


danielli90:
LGTM

skaarthik:
@hebinhuang - can you also add this to build.sh? Looks like Travis CI does not fail when Examples.sln fails to build.

hebinhuang:
I will do it.

PR #405, opened 2016-04-21
renamed mobius config names

skaarthik:
Spark config parsing ignores configs that do not have "spark." prefix


skaarthik:
Code coverage drop is a false alarm as there is no change to C# code or tests

hebinhuang:
LGTM

PR #409, opened 2016-04-22
Fix build.sh to check whether Examples build failed

hebinhuang:
Fix build.sh to check whether Examples build failed


danielli90:
LGTM

hebinhuang:
Code coverage drop is a false alarm as there is no change to C# code or tests

PR #413, opened 2016-04-24
Fix error in csharp/nuget.config

jthelin:
Fix error in csharp/nuget.config file.

Failure mode is:

```
===== Build NuGet package for E:\Depot\GitHub\SparkCLR\csharp\SparkCLR.sln =====
[nugetpack.ps1] [INFO] tagname=[]
NuGet.Config does not contain the expected root element: 'configuration'. 
Path: 'E:\Depot\GitHub\SparkCLR\csharp\NuGet.Config'.
```

PR #404 exposed this issue, but the underlying problem predates that.

### DIFF
--- a/csharp/nuget.config
+++ b/csharp/nuget.config
@@ -1,3 +1,6 @@
-<settings>
-  <repositoryPath>packages</repositoryPath>
-</settings>
+<?xml version="1.0" encoding="utf-8"?>
+<configuration>
+  <settings>
+    <repositoryPath>packages</repositoryPath>
+  </settings>
+</configuration>


skaarthik:
The fix is also in another PR that is failing for a different reason that I have just fixed. About to push the new fix to the existing PR

hebinhuang:
The fix already be merged in another PR

jthelin:
@skaarthik included this fix in PR #410

skaarthik:
@jthelin thanks for the PR. 

BTW, I did not include this fix but fixed the issue (https://github.com/skaarthik/Mobius/commit/10d39571aabdc3183c51d4771f83e9ab7916d2d6) as a part of another PR which was pending on some test failures. I should have followed the best practice and sent a separate PR for the build fix and that could have helped you realize that fix was already in place avoiding wasted cycles.

PR #420, opened 2016-04-25
Copies the dependencies of Examples from runtime/bin to use latest Mobius bits

hebinhuang:
1. Modify the Build.cmd to copy the dependencies of Examples from runtime/bin to use latest Mobius bits.
2. Modify Worker.csproj not to copy App.config, because it will be copied and renamed to CSharpWorker.exe.config automatically.


csharp/Worker/Microsoft.Spark.CSharp/Worker.csproj
```diff
@@ -58,9 +58,7 @@
     </ProjectReference>
   </ItemGroup>
   <ItemGroup>
-    <None Include="App.config">
```
skaarthik: What is this change for?

hebinhuang: The APP.config is always be copied without the change. Actually, APP.config always be copied and renamed to <application>.exe.config by default. So this change is for not coping App.config and only keep CSharpWorker.exe.config.

skaarthik: Got it. LGTM.

skaarthik:
@hebinhuang - I see the following in the build logs. I do not see this error for other examples. Any idea what's happening?

https://ci.appveyor.com/project/SparkCLR/sparkclr/build/1.6.1-SNAPSHOT.732#L787 
mkdir C:\projects\sparkclr\build\examples\Sql\JdbcDataFrame
C:\projects\sparkclr\examples\Sql\JdbcDataFrame\bin\Release\*JdbcDataFrame.* 
The system cannot find the file specified.

https://ci.appveyor.com/project/SparkCLR/sparkclr/build/1.6.1-SNAPSHOT.732#L803
mkdir C:\projects\sparkclr\build\examples\Sql\SparkXml
C:\projects\sparkclr\examples\Sql\SparkXml\bin\Release\*SparkXml.* 
The system cannot find the file specified.

hebinhuang:
Fixed the issue. Please review again - I changed the copy order - copy Mobius bits first, then just copy the new files from Examples

skaarthik:
Will Adapter dll and worker exe from bin get overwritten by the older versions of these files referenced the examples?

hebinhuang:
No. "xcopy /d" copies only those files whose source time is newer than the destination time. in our case, the Adapter dll and worker exe in examples are older than runtime/bin, so these files will not be overwritten.

PR #423, opened 2016-04-26
Avoid unnecessary CSharpRDD if the same partitioner is used in Partit…

qintao1976:
This will reduce one CSharpRDD in UpdateStateByKey stage in SML pipeline.


tawan0109:
LGTM, great discovery.

PR #426, opened 2016-04-26
Fix the exception caused by reference broadcast variables when restor…

qintao1976:
…ing streaming application from checkpoint


xiongrenyi:
should activeSparkContext be internal?

otherwise LGTM

skaarthik:
+1 on making activesparkcontext internal.
@xiongrenyi can you merge this change and send a PR with the update. We cannot wait for @qintao19 if we want to release M2 today

xiongrenyi:
will update to internal and send another PR

PR #406, opened 2016-04-21
Sample word count project in F#.

isaacabraham:

PR #434, opened 2016-05-02
WIP - Fsharp wordcount - trying to fix #406

forki:
trying to fix #406


forki:
Ok the build breaks for some unrelated reason. I wonder if master is building right now

skaarthik:
> Ok the build breaks for some unrelated reason. I wonder if master is building right now

https://github.com/Microsoft/Mobius#build-status. Master branch seems to build fine.

forki:
Yes thanks. I tried that as well.  something else is going on and I hope @isaacabraham and I can fix that.

examples/Batch/WordCountFSharp/WordCountFSharp.fsproj
```diff
@@ -0,0 +1,110 @@
...
+  <ItemGroup>
+    <Compile Include="AssemblyInfo.fs" />
+    <Compile Include="WordCountExample.fs" />
+    <Content Include="App.config" />
+    <Content Include="packages.config" />
+    <None Include="CSharpWorker.exe.config" />
+  </ItemGroup>
+  <ItemGroup>
+    <Reference Include="CSharpWorker">
+      <HintPath>..\packages\Microsoft.SparkCLR.1.5.200\lib\net45\CSharpWorker.exe</HintPath>
```
skaarthik: Shouldn't this be ..\..\packages?

examples/Examples.sln
```diff
@@ -21,6 +21,8 @@ Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Pi", "Batch\Pi\Pi.csproj",
 EndProject
 Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "WordCount", "Batch\WordCount\WordCount.csproj", "{17E4C27F-8441-425A-B82B-23BA5E313CC4}"
 EndProject
+Project("{F2A71F9B-5D33-465A-A702-920D77279786}") = "WordCountFSharp", "Batch\WordCountFSharp\WordCountFSharp.fsproj", "{17B63D32-EFC8-4EF5-831A-197A4FC29F06}"
```
skaarthik: What would be the VS experience for those who do not have F# installed? Consider adding a separate solution for fsproj files if the solution would fail to build without F#.

isaacabraham: An better alternative would be to install F# as part of the installation process, just like already happens with Spark and Java etc. etc.. We want F# to be a first class citizen of the project, that is ready to go rather than an optional add in that people won't see by default.

examples/Examples.sln
```diff
@@ -21,6 +21,8 @@ Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Pi", "Batch\Pi\Pi.csproj",
...
+EndProject
```
forki: Microsoft made VS detect F# projects when you open the solution is opened. It allows you to install the 20ish MB on demand.

examples/Batch/WordCountFSharp/WordCountExample.fs
```diff
@@ -0,0 +1,34 @@
+﻿open Microsoft.Spark.CSharp.Core
+open Microsoft.Spark.CSharp.Services
+open System.Reflection
+open System.Collections.Generic
+
+[<EntryPoint>]
+let main = function
```
dsyme: This use of ``function`` is not particularly idiomatic - in general I think it's better to avoid ``function`` in beginner samples and just build confidence/knowledge with an explicit argument and a ``match``

examples/Batch/WordCountFSharp/WordCountExample.fs
```diff
@@ -0,0 +1,34 @@
...
+    | [| filePath |] ->
+        let logger =
+            LoggerServiceFactory.SetLoggerService Log4NetLoggerService.Instance
+            LoggerServiceFactory.GetLogger (MethodInfo.GetCurrentMethod().DeclaringType)
+        
+        let sparkContext = SparkConf().SetAppName "MobiusWordCount" |> SparkContext
```
dsyme: Any reason that SparkContext isn't an IDisposable?

examples/Batch/WordCountFSharp/WordCountExample.fs
```diff
@@ -0,0 +1,34 @@
...
+        logger.LogInfo (sprintf "Reading from file %s" filePath)
+
+        try
```
dsyme: It feels a shame you need this ceremony of a "main" and a "try"/"catch".  It feels like this whole thing should just be a script (or a .fs in the style of a script).  Speaking of which does this work with F# interactive? thx

examples/Examples.sln
dsyme: @skaarthik I think the experience is that using the project will offer to install F#. Not sure if that's on solution load or project reload.

examples/Batch/WordCountFSharp/WordCountExample.fs
isaacabraham: Not really - you need to create a standalone executable which is then launched from Spark (at least, that's the experience I've had thus far).

Try/catch etc. - again, this is essentially a like-for-like from the C# example so yes it's as much as showing that as anything else.

examples/Batch/WordCountFSharp/WordCountExample.fs
isaacabraham: Good point, will update.

forki:
closing this in favour of #406. @isaacabraham will try to work on @dsyme's feedback

PR #442, opened 2016-05-10
Update download path of winutils.exe to fix running localmode example problem

qualiu:
Update download path of winutils.exe to fix running localmode example problem (such as: permission of /tmp/hive)


qintao1976:
LGTM

skaarthik:
LGTM

PR #443, opened 2016-05-10
Fixing unary operator error in Linux build scripts

slyons:
Empty strings should be checked with `[ -z $foo ]` instead of `[ $foo = "" ]` 🎉


skaarthik:
@slyons - thanks for your contribution. Please sign the CLA using the link above.

PR #444, opened 2016-05-11
Fix error passing arguments of sparkclr-submit to spark-submit when arguments contain special characters.

qualiu:
Fix error passing arguments of sparkclr-submit to spark-submit when arguments contain special characters.

for example : an arg is a connection string as following :
"jdbc:mysql://localhost:3306/lzdb?user=guest&password=abc123"
it contains "&" , cannot pass it from sparkclr-submit.cmd to spark-submit.cmd, even if you quote it like above. 

Error happens in script : https://github.com/Microsoft/Mobius/blob/master/scripts/sparkclr-submit.cmd#L62
The quote (if you added) will disappear in file  LAUNCHER_OUTPUT.

### DIFF
--- a/scala/src/main/org/apache/spark/launcher/SparkCLRSubmitArguments.scala
+++ b/scala/src/main/org/apache/spark/launcher/SparkCLRSubmitArguments.scala
@@ -5,7 +5,8 @@
 
 package org.apache.spark.launcher
 
-import java.io.{PrintStream, File}
+import java.io.{File, PrintStream}
+import java.util.regex.Pattern
 import java.util.{List => JList}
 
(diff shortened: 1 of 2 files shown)


qintao1976:
LGTM

skaarthik:
LGTM

skaarthik:
Closing and reopening to see if "cla-not-required" label is added to the PR

skaarthik:
@qualiu - can you follow-up to find out why your PR has the label "cla-required"?

qualiu:
@skaarthik : Probably because I need signed in https://cla.microsoft.com. (I've just done it)

PR #445, opened 2016-05-11
Add AsyncUnion DStream operation

qintao1976:


xiongrenyi:
1. I am not sure a new API is needed here, it's identical to existing Union semantic wise? starting RDD compute on batch interval is kind of perf tune option, can we use --conf instead?
2. can we simply override existing UnionDStream and RunJob on UnionRDD since what we need is to start it at batch interval even if previous batch not complete?
3. UpdateStateByKey does the same thing now. ssc.progressListener.lastCompletedBatch can be used so that if last batch already completes. no need to start a parallel job.

qintao1976:
1. Here we want a general operation that can transform a normal DStream to "AsyncDStream". The operation should be general enough, not limited to "reading from Kafka" or "UpdateStateByKey".
2. Another reason to add a new operation is that it can combine with other operations. For example, we can begin with an normal kafka direct stream with interval of 1 minute, then transform it to a "AsyncDStream", and then use "Window" operation to transform it to a DStream of 3 minutes batch interval, and finally apply "UpdateByKey". After all these operations, we can achieve the goal that pull from kafka asynchronously in a small batch interval and at the same time "UpdateStateByKey" in a larger interval.
3. I'd prefer not to change the semantic of "UpdateStateBykey" because it is the standard Spark operation. Changing it may surprise our customer and leads to confusion.

xiongrenyi:
1. I agree we should have a more generic API. However, UpdateStateByKey is special in that its 2nd stage must be in sync. i.e. the 2nd stage can't start until previous batch completes. otherwise will run into same issue of concurrentJobs. So if one of union's parent is a UpdateStateByKey (in theory), it will hit the issue.
2. I prefer your Kafka based approach because this approach needs developer to write a different more complex driver program by using a Window at a 1/nth interval of batch interval?
3. UpdateStateByKey's change is on implementation. what do you refer to by 'semantic change'?

skaarthik:
@qintao1976 - I like the idea of generic API for DStream but I am not clear how this PR addresses the problem of batches not starting on-time (and batches getting queued up for execution because of delays)? Is that problem outside the scope of this PR? That is the problem that is addressed in PR #449 in the context of "update state" step. How is this approach of async union better than simply materializing the RDD at the time of creation of CSharpDStream?

qintao1976:
@skaarthik , in this PR, the RDD is not materialized when Compute() method of DStream is called. There are two reasons:
1) Otherwise, it will block streaming UI tab.
2) We want something "async", just like"Receiver mode". In "Receiver Mode", there are two separate things, the first is to receive data from Kafka and create BlockedRDD to cache the received data, the second step is to process the data. In AsyncDStream, RDDs are queued and materialized in separate threads, this is much like what "Receiver" do.
By the way, renyi think the name of operation "unionAsync" is misleading, and I am considering to change the name.

xiongrenyi:
@qintao1976 

I still think we should stick to the original DirectKafkaReceiver or ParallelKafkaDataFetcher:

1. original design has clear 'receiver' semantic, nothing new invented.
2. receiver is IO intensive which totally makes sense to run in parallel, not sure yet extending it to generic async has much value.
3. the approach in question basically requires an extra Window operation which can be confusing to developers and they also needs to understand certain operation like UpdateStateByKey is  not async-able.

skaarthik:
Closing it for now. Feel free to reopen if needed.

PR #446, opened 2016-05-11
Add F# documentation and Maven version data

HankG:
F# documentation to allow F# work count example to compile under build step.  Also version of Maven that is standard in the Debian repository, therefore which is pulled by apt for Debian, Ubuntu, Mint, etc. is 3.0.5 not 3.3.x.  I had no problems running any of the scripts with that version of Maven.


skaarthik:
@HankG Thanks for your contribution

PR #449, opened 2016-05-12
add parallel job to UpdateStateByKey

xiongrenyi:
the parallel job kicks in only when previous batch not complete yet


skaarthik:
We need automated validation. Last time we agreed to take shortcut with manual validation of the sample. I recommend adding unit test now if we cannot get samples to run in Appveyor or Travis.

xiongrenyi:
Scala side unit test? unit test on compute method with RunJob looks much harder. let me try to automate validation in the sample.

xiongrenyi:
added DStreamParallelJobSample with automated validation by using Accumulator to indicate that parallel job has been launched as expected during job generation time. and output test result "Validation passed" or "Validation failed" right above "Successfully finished running sample" message.

skaarthik:
@xiongrenyi - there are three things we need to validate in this change. 
* config switch on & off works
* the value of config is actually used
* the functionality that is implemented in the thread

Is it possible to validate these in the unit test?

xiongrenyi:
I assume you mean Scala side unit test, the compute method in question with runJob needs some mock support to my understanding, which probably worth a new issue.

Due to limited experience on Scala, I definitely need help from the team. I suggest  to create an M3 issue - Scala RDD test framework ? so that the team can jump in to help.

(considering the change itself has been tested lots of times in cluster and don't think anybody outside the team will touch the code soon)

xiongrenyi:
@qintao1976 

How easy do you think to write a unit test for the UpdateStateByKey compute method ?

is it possible to write a framework (or sample code) for us to follow?

Thanks,
Renyi.

xiongrenyi:
unit test failed. investigating...

xiongrenyi:
the unit test failed to access private[streaming] var jobExecutor, but it doesn't happen locally. anybody any idea what could be the reason? 

159[ERROR] C:\projects\sparkclr\scala\src\test\scala\org\apache\spark\streaming\api\csharp\CSharpDStreamSuite.scala:72: error: variable jobExecutor in class CSharpStateDStream cannot be accessed in org.apache.spark.streaming.api.csharp.MockCSharpStateDStream

xiongrenyi:
unit test passed after adding [streaming] scope to jobExecutor private access modifier

Issue #285, opened 2016-01-27
Issue Running PI example on Linux Standalone Cluster


sehunley:
I am able to run the samples locally, but when pointing to my Linux Cluster, I am unable to run the PI sample, as follows:

** Client Mode **
scripts\sparkclr-submit.cmd --proxy-user miadmin --total-executor-cores 2 --master spark://spark01:7077 --exe Pi.exe C:\MyData\Apache_Spark\SparkCLR-master\examples\pi\bin\Debug spark.local.dir %temp% 

Getting the following error(s):
"C:\MyData\Apache_Spark\SparkCLR-master\build\tools\spark-1.6.0-bin-hadoop2.6\conf\spark-env.cmd"
SPARKCLR_JAR=spark-clr_2.10-1.6.0-SNAPSHOT.jar
Zip driver directory C:\MyData\Apache_Spark\SparkCLR-master\examples\pi\bin\Debug to C:\Users\shunley\AppData\Local\Temp\Debug_1453925538545.zip
[sparkclr-submit.cmd] Command to run --proxy-user miadmin --total-executor-cores 2 --master spark://spark01:7077 --name Pi --files C:\Users\shunley\AppData\Local\Temp\Debug_1453925538545.zip --class org.apache.spark.deploy.csharp.CSharpRunner C:\MyData\Apache_Spark\SparkCLR-master\build\runtime\lib\spark-clr_2.10-1.6.0-SNAPSHOT.jar C:\MyData\Apache_Spark\SparkCLR-master\examples\pi\bin\Debug C:\MyData\Apache_Spark\SparkCLR-master\examples\pi\bin\Debug\Pi.exe spark.local.dir C:\Users\shunley\AppData\Local\Temp
[CSharpRunner.main] Starting CSharpBackend!
[CSharpRunner.main] Port number used by CSharpBackend is 4485
[CSharpRunner.main] adding key=spark.jars and value=file:/C:/MyData/Apache_Spark/SparkCLR-master/build/runtime/lib/spark-clr_2.10-1.6.0-SNAPSHOT.jar to environment
[CSharpRunner.main] adding key=spark.app.name and value=Pi to environment
[CSharpRunner.main] adding key=spark.cores.max and value=2 to environment
[CSharpRunner.main] adding key=spark.files and value=file:/C:/Users/shunley/AppData/Local/Temp/Debug_1453925538545.zip to environment
[CSharpRunner.main] adding key=spark.submit.deployMode and value=client to environment
[...]

PR #453, opened 2016-05-17
Fix for issue #285

kai-zeng:
This PR mainly contains two changes for Linux:
1. Fix CSharpWorker path, by adding "./"
2. Preserve file permissions when doing zip & unzip dependencies

### DIFF
--- a/csharp/Adapter/Microsoft.Spark.CSharp/Configuration/ConfigurationService.cs
+++ b/csharp/Adapter/Microsoft.Spark.CSharp/Configuration/ConfigurationService.cs
@@ -114,7 +114,7 @@ namespace Microsoft.Spark.CSharp.Configuration
             /// </summary>
             internal virtual string GetCSharpWorkerExePath()
             {
-                return ProcFileName;
+                return Path.Combine(".", ProcFileName);
             }
         }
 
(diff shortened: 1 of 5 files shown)


kai-zeng:
BTW, CSharpWorker.exe downloaded from nuget does not have executable permission on Linux. We need to fix this @skaarthik

skaarthik:
@kai-zeng we can fix the permissions of CSharpWorker.exe in NuGet before the next release. For the past releases, I am wondering if we could have a special check in place just for CSharpWorker.exe at the time of unzipping and set the write permissions if needed. I believe that would take care of YARN mode execution in Linux. For local and standalone modes in Linux, where zipping/unzipping is not done, we could document the requirement to update the permissions of CSharpWorker.exe in the resources folder before app submission.

kai-zeng:
ping... 
@skaarthik 
@tawan0109

kai-zeng:
@skaarthik Sure. We can add that check to branch 1.5 and 1.6. I think for master we don't that check

skaarthik:
@kai-zeng - example projects in master branch point to the last release (1.6.100) in NuGet that does not have the right permissions for CSharpWorker.exe. So, I am not sure why you think the check is not necessary in master branch. Am I missing something?

kai-zeng:
@skaarthik I thought example projects in our master branch is referencing a daily snapshot package. If it references release 1.6.1, then we do need the check.
BTW, shouldn't example projects reference a daily built package?

skaarthik:
@kai-zeng NuGet is not good for publishing daily snapshot builds. Guan was exploring myget.org for that. I do not know where it landed. We were also hoping that example projects would serve as templates for developers to start building their Mobius apps.

It is a topic that needs further discussion. So, let's continue this conversation outside the PR. I hope you are fine with the idea of updating permissions in master also along with other branches.

kai-zeng:
@skaarthik Sure. I will add the check.

skaarthik:
Thanks @kai-zeng 
LGTM

PR #457, opened 2016-05-19
Fixing issue that build did not stop when mvn build failed

tawan0109:
* `copy` operation overwrites `%ERRORLEVEL%` returned by `mvn.cmd build`, which caused build script failed to check whether scala build succeed or not.

* This PR tries to saves `%ERRORLEVEL%` to a variable `MVN_ERRORLEVEL` before do any other operations, and use this variable to check mvn build status.


skaarthik:
LGTM

PR #458, opened 2016-05-20
fix Duration type mismatch bug in Window API and enable related sample

xiongrenyi:


xiongrenyi:
1. exclude DStreamTextFileSample from build by reverting back to experimental
2. include Window() API in question into unit test
3. create a Duration proxy method instead of calling JVM Duration directly from everywhere

danielli90:
LGTM. Suggest to add a comment to justify the switch to seconds.

xiongrenyi:
this PR changes duration to seconds instead of mini seconds in streaming batch interval to be consistent with Checkpoint API, Remember API and Windows API like python does.
also refactor to have a Duration method instead of calling JVM Duration directly from everywhere

Issue #460, opened 2016-05-25
failure when submitting Mobius app from Windows client to Spark cluster in Linux


skaarthik:
Two issues that need to be fixed:
* Paths.get() implementation is Java is OS-specific. So we need an extra configuration parameter to indicate that the OS from which the job submitted is different than the OS in which Spark is running. That flag need to be used at https://github.com/Microsoft/Mobius/blob/master/scala/src/main/org/apache/spark/api/csharp/CSharpRDD.scala#L50
* When the job is submitted from Windows, the permissions on the artifacts submitted to the cluster is 000. So the driver exe cannot be executed. Need to update permission of all files in the zip folder to 777 (or something else?) at the time of unzip.

@kai-zeng - comments?

kai-zeng:
- As to the path configuration, I think we don't need the extra configuration parameter. Instead, I prefer to construct the path in CSharpRDD.compute. That is, in C# spark context codes, we simply set `cSharpWorkerExecutable = "CSharpWorker.exe"`, and in CSharpRDD.compute, we construct path using `Path.get(".", cSharpWorkerExecutable)`.
There are two reasons for this approach:
1. compute is happening on the spark worker nodes, and this is the place we actually use the path. Constructing the path here is guaranteed to work no matter whether the client and the cluster are using the same type of OS. 
2. PythonRDD.compute simply calls PythonRunner, as shown below.
`val runner = new PythonRunner(
      command, envVars, pythonIncludes, pythonExec, pythonVer, broadcastVars, accumulator,
      bufferSize, reuse_worker)
    runner.compute(firstParent.iterator(split, context), split.index, context)`
Overriding PythonRDD.compute to our path construction requires minimal codes. 

- As to file permissions, I suggest that we always set the unix modes for all files in the zip file. 
When zipping dependencies: 
1. if on linux, we simply use the file permissions of the original files
2. if on windows,we set the file permissions to 755
When unzipping dependencies:
1. if on linux, we set the file permissions according the unix modes in the zip file
2. if on windows, we simply ignore the unix modes in the zip file

skaarthik:
That would work too. Avoiding a new config is better. We already override compute method. I assumed you had a reason for not using compute method to set/correct worker path in your recent commit.

Suggestion on setting permissions sounds good.

kai-zeng:
Actually for the recent commit I don't have a strong reason against overriding compute. I just wanted to stick to the current approach, but I didn't consider submitting jobs across different OS.

[...]

PR #461, opened 2016-05-26
Fix for issue #460

kai-zeng:
1. construct csharp worker path in CSharpRDD.compute; 
2. set unix mode in zip files for cross-platform supports


scala/src/main/org/apache/spark/api/csharp/CSharpRDD.scala
```diff
@@ -90,7 +72,11 @@ class CSharpRDD(
       logInfo(s"workerFactoryId: $workerFactoryId")
     }
 
-    super.compute(split, context)
+    val cSharpExec = cSharpWorker.getAbsolutePath
```
skaarthik: Looks like this could be removed

skaarthik:
LGTM

PR #462, opened 2016-05-26
fix DynamicPartitionKafkaInputDStream checkpint and add C# func support

xiongrenyi:
1. fix DynamicPartitionKafkaInputDStream checkpoint by fixing restore method and starting streaming callback server ealier during restore time
2. add C# func support in DynamicPartitionKafkaInputDStream by inheriting from InputDStream directly

### DIFF
--- a/scala/src/main/org/apache/spark/streaming/api/csharp/CSharpDStream.scala
+++ b/scala/src/main/org/apache/spark/streaming/api/csharp/CSharpDStream.scala
@@ -102,8 +102,8 @@ class CSharpDStream(
   override def slideDuration: Duration = parent.slideDuration
 
   override def compute(validTime: Time): Option[RDD[Array[Byte]]] = {
-    val rdd = parent.compute(validTime)
-    if (rdd.isDefined) {
+    val rdd = parent.getOrCompute(validTime)
+    if (rdd.isDefined && cSharpFunc != null && !cSharpFunc.isEmpty) {
       CSharpDStream.callCSharpTransform(List(rdd), validTime, cSharpFunc, List(serializationMode))
     } else {
       None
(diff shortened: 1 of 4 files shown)


scala/src/main/org/apache/spark/streaming/api/csharp/CSharpDStream.scala
```diff
@@ -102,7 +102,7 @@ class CSharpDStream(
   override def slideDuration: Duration = parent.slideDuration
 
   override def compute(validTime: Time): Option[RDD[Array[Byte]]] = {
-    val rdd = parent.compute(validTime)
+    val rdd = parent.getOrCompute(validTime)
```
skaarthik: What is the reason for this change?

scala/src/test/scala/org/apache/spark/streaming/api/kafka/DynamicPartitionKafkaInputDStreamSuite.scala
```diff
@@ -0,0 +1,82 @@
...
+package org.apache.spark.streaming.kafka
+
+import kafka.common.TopicAndPartition
+import kafka.message.MessageAndMetadata
+import kafka.serializer.DefaultDecoder
+import org.apache.spark.csharp.SparkCLRFunSuite
+import org.apache.spark.streaming.{Duration, StreamingContext, Time}
+import org.apache.spark.{Partition, SparkConf, SparkContext}
+
+class DynamicPartitionKafkaInputDStreamSuite extends SparkCLRFunSuite {
```
skaarthik: Good to see a new test suite. Can you also add test for compute method?

scala/src/main/org/apache/spark/streaming/api/csharp/CSharpDStream.scala
xiongrenyi: it's a bug I think.
getOrCompute put generated RDDs in a HashMap for future reference including checkpoint.
double checked python did same thing.
not sure why it was compute.

skaarthik: Ok, missing test is probably one of the reasons for this bug left uncovered. Can we add a test now?

xiongrenyi:
Added by calling CSharpDStream.compute

scala/src/test/scala/org/apache/spark/streaming/api/kafka/DynamicPartitionKafkaInputDStreamSuite.scala
xiongrenyi: Added

qintao1976:
LGTM

PR #463, opened 2016-05-27
Fix AwaitTermination(int timeout) error

qualiu:
Fix AwaitTermination(int timeout) error, and change to AwaitTerminationOrTimeout(long timeout).


qintao1976:
LGTM

PR #464, opened 2016-05-28
Fix for issue #460 in Branch 1.6

kai-zeng:
1. construct csharp worker path in CSharpRDD.compute;
2. set unix mode in zip files for cross-platform supports


skaarthik:
LGTM

PR #466, opened 2016-05-31
support parallelly pre-fetch data for DynamicPartitionKafkaInputDStream

qintao1976:
Add ParallelKafkaDataFetcherDStream to fetch data from Kafka in parallel

### DIFF
--- a/scala/src/main/org/apache/spark/streaming/api/csharp/CSharpDStream.scala
+++ b/scala/src/main/org/apache/spark/streaming/api/csharp/CSharpDStream.scala
@@ -5,16 +5,19 @@
 
 package org.apache.spark.streaming.api.csharp
 
-import org.apache.spark.api.csharp._
-import org.apache.spark.api.csharp.SerDe._
+import scala.collection.JavaConversions._
+import scala.collection.mutable.ListBuffer
+import scala.language.existentials
+
 import java.io.DataInputStream
 import java.io.DataOutputStream
 import java.net.Socket
 import java.util.{ArrayList => JArrayList}
-import java.util.concurrent.ThreadPoolExecutor
+import java.util.concurrent.{LinkedBlockingQueue, ConcurrentHashMap, ThreadPoolExecutor}
 
-import scala.collection.JavaConversions._
-import scala.language.existentials
+import org.apache.spark.{Logging, SparkContext}
+import org.apache.spark.api.csharp._
+import org.apache.spark.api.csharp.SerDe._
 import org.apache.spark.api.java._
 import org.apache.spark.rdd._
 import org.apache.spark.storage.StorageLevel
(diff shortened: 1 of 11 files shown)


xiongrenyi:
@qintao1976 
1. the goal here is to compute KafkaRDD asynchronously with KafkaDStream compute? if we compute KafkaRDD in setOffsetsRangeForNextBatch method, will it achieve the same?
2. the new DStream should be same as existing DynamicPartitionKafkaInputDStream except that its KafkaRDD gets materialized?

qintao1976:
Yes, the basic idea is simple. But the implementation is a bit complicated due to the following factors:
1) Because the memory of executors is limited, we must control how many RDDs can be materialized in advance. That's to say, we need some backpressure machanism.
2) The interval of refresh offsets may be different from the batch interval.

xiongrenyi:
@qintao1976 

1. We fixed KafkaRDD checkpoint issue and added C# func support, why still need a new DStream?
2. Can backpressure be simply implemented on pre-computed RDD queue size? i.e. if RDD queue size > config value ...

xiongrenyi:
@qintao1976 

I mean your original implementation of asynchronous metadata refresh already has the right structure. what we want is to further generate / materialize the rdd as well? i.e.

    refreshOffsetsScheduler.scheduleAtFixedRate(new Runnable {
      override def run(): Unit = {
        setOffsetsRangeForNextBatch()
        if (numReceivers > 0 && (maxNumReceivers == 0 || rddQueue.size < maxNumReceivers)) {
          val rdd = computeRDD(new Time(0)).get.cache() // existing compute but without reportInfo
          jobExecutor.execute(new Runnable {
            override def run(): Unit = {
              context.sparkContext.runJob(rdd, (iterator: Iterator[R]) => {})
              rddQueue.enqueue(rdd)
            }})}}
    }, 0, refreshOffsetsInterval, TimeUnit.MILLISECONDS)

and refactor compute, update and restore methods a little bit should be good enough?

xiongrenyi:
@qintao1976 @tawan0109  @skaarthik 

1. articulate the above as a proposal in codeflow - comments?

codeflow:open?server=http://codeflow/Services/DiscoveryService.svc&review=renyix-a31d559a7416462f945169d95ef2ec96&alert=true

2. Do we really need a new API CreateParallelKafkaDataFetcherDStream here? is it different from CreateDirectStream from user's perspective? since the only difference here to my understanding is implementation?

xiongrenyi:
API name from user perspective ? or use existing API with --conf command line option?

Otherwise LGTM

scala/src/main/org/apache/spark/streaming/api/csharp/CSharpDStream.scala
```diff
@@ -343,3 +347,149 @@ class CSharpConstantInputDStream(ssc_ : StreamingContext, rdd: RDD[Array[Byte]])
...
+          record.rdd.persist(storageLevel)
+          sc.setCallSite(s"runJob for [$id]")
+          sc.runJob(record.rdd, (iterator: Iterator[T]) => {})
+        }
+      })
+    }
+  }
+
+  // input RDD
+  def slurp(userInfo: Any, rdd: RDD[T]): Unit = {
```
skaarthik: Just wondering why you are using slurp and spit instead of put and get that are common terminologies used with cache implementations

scala/src/main/org/apache/spark/streaming/api/kafka/KafkaOffsetsFetcher.scala
```diff
@@ -0,0 +1,241 @@
...
+
+/**
+ * reading metadata of multiple topics from across multiple data centers takes long time to
+ * complete, which impacts DStream performance and causes UI steaming tab not responsive due to
+ * mutex held by DStream, so a separate thread is introduced to refresh metadata (current offsets)
+ * asynchronously at below interval. This unblocks DStream in above described situation but not
+ * quite in sync with batch timestamp, which is OK since batches are still generated at the same
+ * interval
+ */
+class KafkaOffsetsFetcher(
```
skaarthik: we need tests for this class

scala/src/main/org/apache/spark/streaming/api/kafka/ParallelKafkaDataFetcherDStream.scala
```diff
@@ -0,0 +1,270 @@
...
+    topics: Set[String],
+    val fromOffsets: Map[TopicAndPartition, Long],
+    messageHandler: MessageAndMetadata[K, V] => R,
+    numPartitions: Int,
+    cSharpTransformFunc: Array[Byte]
+  ) extends InputDStream[R](_ssc) {
+
+  private val serializationMode = "Pair"
+
+  private val refreshOffsetsInterval = context.sparkContext.getConf.getLong(
```
skaarthik: documentation updates needed for new config parameters @ https://github.com/Microsoft/Mobius/blob/master/notes/configuration-mobius.md

scala/src/main/org/apache/spark/streaming/api/csharp/CSharpDStream.scala
```diff
@@ -343,3 +347,149 @@ class CSharpConstantInputDStream(ssc_ : StreamingContext, rdd: RDD[Array[Byte]])
...
+
+case class RddPreComputeRecord[T] (
+    rddSeqNum: Long,
+    userInfo: Any,
+    rdd: RDD[T])
+
+/*
+ * Slurp RDDs, apply cache and pre-compute operation asynchronously, then spit them out
+ */
+class RddPreComputeCache[T](
```
skaarthik: Consider renaming. Cache implementations typically do not operate on the data. That is not the case with this implementation. Suggestion - RddPreComputeProcessor

scala/src/main/org/apache/spark/streaming/api/csharp/CSharpDStream.scala
```diff
@@ -343,3 +347,149 @@ class CSharpConstantInputDStream(ssc_ : StreamingContext, rdd: RDD[Array[Byte]])
...
+    while (true) {
+      val record = pendingRddQueue.take()
+      logInfo(s"process pending queue [$id], rddSeqNum: ${record.rddSeqNum}, " +
+        s"ackedSeqNum: $ackedSeqNum, outputRddSeqNum: $outputRddSeqNum")
+      while (isPending(record.rddSeqNum)) {
+        Thread.sleep(100)
+      }
+      logInfo(s"process pending queue [$id], submitting job, rddSeqNum: ${record.rddSeqNum}, " +
+        s"ackedSeqNum: $ackedSeqNum, outputRddSeqNum: $outputRddSeqNum")
+      jobExecutor.execute(new Runnable {
```
skaarthik: Consider extracting to a method and reuse between this usage and the usage in slurp()

scala/src/main/org/apache/spark/api/csharp/CSharpRDD.scala
```diff
@@ -65,7 +65,7 @@ class CSharpRDD(
       command(i) = bytes(i)
     }
 
-    if (CSharpRDD.maxCSharpWorkerProcessCount >= 0) {
```
skaarthik: Could this have been caught earlier with unit tests?

scala/src/main/org/apache/spark/streaming/api/csharp/CSharpDStream.scala
```diff
@@ -343,3 +347,149 @@ class CSharpConstantInputDStream(ssc_ : StreamingContext, rdd: RDD[Array[Byte]])
...
+      }
+    }
+
+    if (result.isEmpty) {
+      stop = false
+      spitNum = 0
+      while (!stop) {
+        val rddSeqNum = outputRddSeqNum
+        if (bypassRddMap.containsKey(rddSeqNum)) {
+          result += bypassRddMap.remove(rddSeqNum)
```
skaarthik: So, in this case, RDD is not precomputed right? (but it is still passed under the datastructure named RddPreComputeRecord?)

scala/src/main/org/apache/spark/streaming/api/kafka/ParallelKafkaDataFetcherDStream.scala
```diff
@@ -0,0 +1,270 @@
...
+    }
+
+    override def update(time: Time) {
+      batchForTime.clear()
+      for (t <- preComputeOutputMap.keySet().toList) {
+        if (generatedRDDs.contains(t)) {
+          batchForTime +=
+            (t -> preComputeOutputMap.get(t).map(_.userInfo.asInstanceOf[KafkaOffsetsFetchResult]))
+        } else {
+          preComputeOutputMap.remove(t)
```
skaarthik: Can you explain this condition - when it happens and why you are removing from the map?

scala/src/main/org/apache/spark/streaming/api/kafka/ParallelKafkaDataFetcherDStream.scala
```diff
@@ -0,0 +1,270 @@
...
+    }
+
+    if (offsetsScheduler != null) {
+      offsetsScheduler.shutdown()
+      offsetsScheduler = null
+    }
+  }
+
+  private[streaming]
+  class ParallelKafkaDataFetcherCheckpointData extends DStreamCheckpointData(this) {
```
skaarthik: Great to see that you are extending checkpoint data and making sure the implementation of the new DStream is robust

scala/src/test/scala/org/apache/spark/streaming/api/kafka/ParallelKafkaDataFetcherDStreamSuite.scala
```diff
@@ -0,0 +1,92 @@
...
+import org.apache.spark.rdd.RDD
+import org.apache.spark.streaming.api.csharp.{RddPreComputeRecord, CSharpDStream}
+import org.apache.spark.streaming.kafka.KafkaCluster.LeaderOffset
+import org.apache.spark.streaming.scheduler.InputInfoTracker
+import org.apache.spark.streaming.{Duration, StreamingContext, Time}
+import org.apache.spark.{Partition, SparkConf, SparkContext}
+
+import scala.collection.mutable.ListBuffer
+
+class ParallelKafkaDataFetcherDStreamSuite extends SparkCLRFunSuite {
```
skaarthik: Thanks for adding these tests

skaarthik:
+1 on the suggestion from @xiongrenyi on adding this functionality to the existing API.

scala/src/main/org/apache/spark/streaming/api/csharp/CSharpDStream.scala
qintao1976: 1. Yes, put and get are better, and I will change the name. BTW, Slupr and spit are names from clojure language:
https://clojuredocs.org/clojure.core/slurp
https://clojuredocs.org/clojure.core/spit

scala/src/main/org/apache/spark/streaming/api/kafka/KafkaOffsetsFetcher.scala
qintao1976: I will have a try. But the UNIT test of this class involves Kafka and ZooKeeper and may be difficult to do.

scala/src/main/org/apache/spark/streaming/api/kafka/ParallelKafkaDataFetcherDStream.scala
qintao1976: Will update the doc.

scala/src/main/org/apache/spark/streaming/api/csharp/CSharpDStream.scala
qintao1976: Thanks for the suggestion. Will change the name.

scala/src/main/org/apache/spark/streaming/api/csharp/CSharpDStream.scala
qintao1976: Will consider the suggestion and refactor the code.

scala/src/main/org/apache/spark/api/csharp/CSharpRDD.scala
qintao1976: No. In the original design, we think there is no concurrent job running in the application. But pre-fetch changes this assumption. So this changes the implementation of long running C# process..

scala/src/main/org/apache/spark/streaming/api/csharp/CSharpDStream.scala
qintao1976: Yes. Backpressure may cause the RDD not to be precomputed. But the job generator  thread is still running and create new job for every batch interval. The precomputation of "bypassed" RDD will be triggered when backpressure condition quits.

qintao1976:
Thanks for all the comments. I think the code need some refactor to accommodate Renyi and Kaarthik's review comments. Will do that next week.

qintao1976:
Merge this PR because it is pending for a very long time. Perhaps there is still something that can be further improved, let us do that later.

Issue #468, opened 2016-05-31
error launching worker in CentOS and OSX


skaarthik:
In CentOS and OSX, to execute Mobius either "binfmt_misc" has to be used or the [recommendation](http://www.mono-project.com/docs/getting-started/application-deployment/#layout-recommendation) of creating a script to call the application needs to be employed. Using the script for Mobius driver applications is straightforward but there is no way for doing that for CSharpWorker. 

Here is simple way to fix the problem:
* We will release CSharpWorker.sh.exe that will just point to CSharpWorker.exe with "mono" prefix (Mono path in this script needs to be configured by the user) 
* User will set the value of CSharpWorkerPath in the app.config of the driver application (similar to the debug mode) to "CSharpWorker.sh,exe" 

@qualiu & @kai-zeng - comments?

qualiu:
@skaarthik - That's a good idea！I agree that leave trivial works  as little as possible to user.

kai-zeng:
I have a slight different suggestion: instead of adding a CSharpWorker.sh.exe, what about we prefix the path with "mono" when we constructing csharpworker path in CSharpRDD.compute?

skaarthik:
Adding "mono" will not be enough. I had to do the following in CSharpWorker.sh.exe to get it to work.
export PATH=$PATH:/opt/mono/bin
export PKG_CONFIG_PATH=/opt/mono/lib/pkgconfig

Adding these export statements to a script at /etc/profile.d and restarting Spark services did not help.

Maybe there is another way to set the right path and pkg config and in that case we could just prefix "mono" to worker path.

qualiu:
First, I think it's better to solve it at the beginning:
1) Modifying the system once for all is not a best idea. (maybe that's why http://www.mono-project.com/docs/getting-started/application-deployment/#layout-recommendation)
2) Temporally changing as following is annoying and may impossible for user who can't get root permission :
echo ':CLR:M::MZ::/usr/bin/mono:' > /proc/sys/fs/binfmt_misc/register

So I think adding CSharpWorker.sh.exe is good idea.

Secondly, for the detail , I think it's better do it in an automatic way. Take mono path as example:
1) Try get it from user settings , like environment settings at first.
2) Automatically detect and get the mono-PATH (like "locate mono" or "whereis mono" ),
3) If failed to find mono, then asking user to provide the path.

skaarthik:
@kai-zeng - any additional comments?

kai-zeng:
[...]

Issue #467, opened 2016-05-31
YARN cluster mode failure in Linux


skaarthik:
Client mode works fine in Linux but cluster mode fails with the message like the following:
CSharpRunner.main]Cannot run program "SparkClrPi.exe": error=2, No such file or directory

We may have to add "." to the exe before launching it as a process in CSharpRunner.

@kai-zeng & @tawan0109 - FYI

kai-zeng:
Agreed.

skaarthik:
Fixed with https://github.com/Microsoft/Mobius/commit/4004001732470d619633ee311fd7e9ca56f79b8a

PR #470, opened 2016-06-04
adding instructions for Linux

skaarthik:
This is just the initial draft of doc. It needs to be updated once #467 and #468 are fixed


skaarthik:
any comments before I sqash and merge the changes?

PR #471, opened 2016-06-10
Cassandra example and other minor updates

skaarthik:


skaarthik:
ping...

examples/Examples.sln
```diff
@@ -1,7 +1,6 @@
-
 Microsoft Visual Studio Solution File, Format Version 12.00
-# Visual Studio 2013
-VisualStudioVersion = 12.0.30501.0
+# Visual Studio 14
```
danielli90: Visual Studio "14" corresponds to Visual Studio 2015. To simplify supporting matrix, i think it is totally fine that we state that VS 2015 and above is required. If that is the direction we are heading to, we just need to update build notes - https://github.com/Microsoft/Mobius/blob/master/notes/windows-instructions.md#building-mobius.

notes/running-mobius-app.md
```diff
@@ -180,6 +180,18 @@ Displays the number of XML elements in the input XML file provided as the first
 `sparkclr-submit.cmd --jars <jar files used for using Hive in Spark> --exe HiveDataFrame.exe C:\Git\Mobius\examples\Sql\HiveDataFrame\bin\Debug`
 
 Reads data from a csv file, creates a Hive table and reads data from it
+
+### Cassandra Example (Sql)
+* Download following jars that are needed to use Spark with Cassandra. **Note** that you need to get the right version of the jar files depending on the versions of Spark and Cassandra.
```
danielli90: It will be nice to add a URL pointer to Cassandra connnect tutorial/wiki page as a reference, so the reader can go to Cassandra Connector page to read about how to include Cassandra jars or download Cassandra spark packages (the reader may find there is a spark-package option for leveraging Cassandra connector).

examples/Sql/CassandraDataFrame/Program.cs
```diff
@@ -0,0 +1,90 @@
...
+{
+    /// <summary>
+    /// This example shows how to use Cassandra in .NET Apache Spark DataFrame API
+    /// It covers load rows from Cassandra, filter it and save results to another table
+    /// </summary>
+    public class CassandraDataFrameExample
+    {
+        static void Main(string[] args)
+        {
+            var cassandraHostName = "localhost";
```
danielli90: It will be nice to the option of passing cassandra hostname/keyspace/table through arguments (or app.config).

danielli90:
Added a few minor comments above. VS version consistency is important. Other than that, LGTM.

PR #474, opened 2016-06-14
Release JVM object reference in Java(Scala) side JVMObjectTracker from C# side.

qualiu:
Release JVMObjectTracker oject reference.
1.Java-side: https://github.com/Microsoft/Mobius/blob/master/scala/src/main/org/apache/spark/api/csharp/CSharpBackendHandler.scala#L269
JVMObjectTracker keep a HashMap[String, Object] which is [id, Java-object]

2.CSharp-side :
    1) JvmObjectReference remember the id : https://github.com/Microsoft/Mobius/blob/master/csharp/Adapter/Microsoft.Spark.CSharp/Interop/Ipc/JvmObjectReference.cs#L20 
    2) So JvmBridge can call java object's method https://github.com/Microsoft/Mobius/blob/master/csharp/Adapter/Microsoft.Spark.CSharp/Interop/Ipc/JvmBridge.cs#L69
    So potential memory leak can happen in JVMObjectTracker.
    To solve this, track the garbage collection in CSharp side, get the id, release JVMObjectTracker's HashMap.

### DIFF
--- a/csharp/Adapter/Microsoft.Spark.CSharp/Interop/Ipc/JvmObjectReference.cs
+++ b/csharp/Adapter/Microsoft.Spark.CSharp/Interop/Ipc/JvmObjectReference.cs
@@ -21,6 +21,7 @@ namespace Microsoft.Spark.CSharp.Interop.Ipc
         {
             Id = jvmReferenceId;
             creationTime = DateTime.UtcNow;
+            WeakObjectManager.GetWeakObjectManager().AddWeakRefereceObject(this);
         }
 
         public override string ToString()
--- a/csharp/Adapter/Microsoft.Spark.CSharp/Adapter.csproj
+++ b/csharp/Adapter/Microsoft.Spark.CSharp/Adapter.csproj
@@ -77,6 +77,7 @@
     <Compile Include="Core\StorageLevel.cs" />
     <Compile Include="Interop\Ipc\JsonSerDe.cs" />
     <Compile Include="Interop\Ipc\JvmBridgeUtils.cs" />
+    <Compile Include="Interop\Ipc\WeakObjectManager.cs" />
     <Compile Include="Interop\SparkCLREnvironment.cs" />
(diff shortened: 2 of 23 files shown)


skaarthik:
As suggested in codeflow, either make SourceLinesSocket, testArgsQuotes and testKeyValueStream a part of existing unit tests to include with build automation or not include them at all in the PR.

csharp/Adapter/Microsoft.Spark.CSharp/Interop/Ipc/JvmObjectReference.cs
```diff
@@ -21,6 +21,7 @@ public JvmObjectReference(string jvmReferenceId)
         {
             Id = jvmReferenceId;
             creationTime = DateTime.UtcNow;
+            WeakObjectManager.GetWeakObjectManager().AddWeakRefereceObject(this);
```
skaarthik: GetWeakObjectManager() does not have any other reference, So, would it be better to delete that method and make AddWeakRefereceObject() a internal static method that internally uses the static member WeakObjManager directly?

qualiu: Then to remove interface IWeakObjectManager or modify it ?

skaarthik: Oh, I missed the interface. I think it would help to keep the interface and use it in unit tests. Add unit tests and see if you can remove the interface. If you can, go ahead and remove it to implement the suggestion above. If you cannot remove the interface, without impacting unit tests, just keep it.

qualiu: I updated the code review. please add your comment.

danielli90:
Thank @hebinhuang for suggesting WeakReference class, @qualiu for implementing, and @skaarthik for code review. This is a very critical fix.

PR #475, opened 2016-06-14
Fix build scripts to comsume msbuild v12.0

hebinhuang:
Currently, we are using the msbuild with v4.0 to compile .NET code. this change is to consume v12.0 one.

### DIFF
--- a/csharp/Build.cmd
+++ b/csharp/Build.cmd
@@ -5,11 +5,13 @@ SET CMDHOME=%~dp0
 @REM Remove trailing backslash \
 set CMDHOME=%CMDHOME:~0,-1%
 
-@REM Set some .NET directory locations required if running from PowerShell prompt.
-if "%FrameworkDir%" == "" set FrameworkDir=%WINDIR%\Microsoft.NET\Framework
-if "%FrameworkVersion%" == "" set FrameworkVersion=v4.0.30319
+@REM Set msbuild location.
+SET VisualStudioVersion=12.0
 
-SET MSBUILDEXEDIR=%FrameworkDir%\%FrameworkVersion%
+SET MSBUILDEXEDIR=%programfiles(x86)%\MSBuild\%VisualStudioVersion%\Bin
+if NOT EXIST "%MSBUILDEXEDIR%\." SET MSBUILDEXEDIR=%programfiles%\MSBuild\%VisualStudioVersion%\Bin
+if NOT EXIST "%MSBUILDEXEDIR%\." GOTO :ErrorMSBUILD
+
 SET MSBUILDEXE=%MSBUILDEXEDIR%\MSBuild.exe
 SET MSBUILDOPT=/verbosity:minimal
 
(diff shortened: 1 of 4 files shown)


skaarthik:
What is the reason for this change?

hebinhuang:
The version of msbuild consumed in existing script is 4.6.1038.0 (under C:\Windows\Microsoft.NET\Framework), which is very old. From VS2013, the default path of MSBUILD is changed to "C:\Program Files (x86)\MSBuild\12.0\bin\". You can get the path from your machine's regedit: HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\MSBuild\ToolsVersions\12.0\.
 Currently, we are officially support Visual Studio 2013 and later, so we should change the msbuild version to match it.

Another reason to change is that the msbuild 4.0 does not support C++ 11. my native code cannot be compiled.

csharp/Clean.cmd
```diff
@@ -1,3 +1,3 @@
-FOR /F "tokens=*" %%G IN ('DIR /B /AD /S bin') DO RMDIR /S /Q "%%G"
```
skaarthik: Can you explain what's changing here? Also consider adding REM comments to the batch script for folks (like me) who find it hard to read & understand the script

skaarthik:
LGTM other than the minor comment on adding comments to batch scripts

xiongrenyi:
LGTM

csharp/Clean.cmd
hebinhuang: the existing clean.cmd just delete sub-folders and files under "bin" and "obj", but it do not delete 'bin' and 'obj' folder self. So after running clean.cmd, you still can see these two folders that we want to clean up.

csharp/Build.cmd
```diff
@@ -5,11 +5,12 @@ SET CMDHOME=%~dp0
...
 set CMDHOME=%CMDHOME:~0,-1%
 
-@REM Set some .NET directory locations required if running from PowerShell prompt.
-if "%FrameworkDir%" == "" set FrameworkDir=%WINDIR%\Microsoft.NET\Framework
-if "%FrameworkVersion%" == "" set FrameworkVersion=v4.0.30319
+@REM Set msbuild location.
+SET VisualStudioVersion=12.0
+
+SET MSBUILDEXEDIR=%programfiles(x86)%\MSBuild\%VisualStudioVersion%\Bin
+if NOT EXIST "%MSBUILDEXEDIR%\." SET MSBUILDEXEDIR=%programfiles%\MSBuild\%VisualStudioVersion%\Bin
```
skaarthik: what if this path also does not exist? Can we display an error message about missing msbuild and goto eof?

hebinhuang: Sure, let me add it.

hebinhuang: Added

danielli90:
LGTM

PR #478, opened 2016-06-20
Fix comment line prefix in run_samples.cmd script.

jthelin:


qualiu:
@jthelin : Nice catch !   Good to Me.
The [compilation error](https://s3.amazonaws.com/archive.travis-ci.org/jobs/138907621/log.txt)  seems due to some Ubuntu software package installation and NuGet restore failure (timeout).
Probably re-run it will come back to normal.

danielli90:
LGTM.

PR #479, opened 2016-06-21
Disabled LogFileAppender for CSharpWorker

tawan0109:
* Usually it's not necessary to enable two log appenders at the same time. For `LogFileAppender`, the output location is under `%TEMP%` directory (on drive `C:\` by default). We should let user to make the call whether to write logs to `%TEMP%` directory.
* Comment out `LogFileAppender` in CSharpWorker `app.config`, only `ConsoleAppender` is enabled by default

### DIFF
--- a/csharp/Worker/Microsoft.Spark.CSharp/App.config
+++ b/csharp/Worker/Microsoft.Spark.CSharp/App.config
@@ -11,7 +11,9 @@
     <root>
       <level value="INFO" />
       <appender-ref ref="ConsoleAppender" />
+      <!--
       <appender-ref ref="LogFileAppender" />
+      -->
     </root>
     <appender name="ConsoleAppender" type="log4net.Appender.ConsoleAppender">
       <layout type="log4net.Layout.PatternLayout">


csharp/Worker/Microsoft.Spark.CSharp/App.config
```diff
@@ -11,7 +11,9 @@
     <root>
       <level value="INFO" />
       <appender-ref ref="ConsoleAppender" />
+      <!--
       <appender-ref ref="LogFileAppender" />
+      -->
```
qualiu: Why not remove it to keep clean and less code ? We've git history.

tawan0109: I considered removing this line directly. 
However, IMO, if we comment it out instead of deleting it, it will be much easier for Mobius users to notice that there is an OOTB `LogAppender` which writes logs to files. If users use nuget package, there is no git history available.

qualiu: Ok, that's fine.

skaarthik:
LGTM

@tawan0109 we do not have any documentation on configuring logs for CSharpWorker...would you mind adding a section to https://github.com/Microsoft/Mobius/blob/master/notes/configuration-mobius.md explaining this?

tawan0109:
@skaarthik Sure, I will send another PR to cover this piece.

PR #483, opened 2016-06-24
Add RIOSocketWrapper and SaeaSocketWrapper

hebinhuang:
Run Sparkclrsperf.exe for perf suite RunRDDMaxDeletionsByUser only (because this suite invoke CSharpWorker).

There is 11% perf improvement by using RIO socket for the file with 4KB a line, and 15% perf improvement for the file with 64KB a line.

Traditional Socket for file with 4KB a line:
User with max deletions is /user/mwcl_musicbrainz, count of deletions=237776. Elapsed time=00:04:35.1505607
Requesting to close all call back sockets.
** Printing results of the perf run (C#) **
** Execution time for RunRDDMaxDeletionsByUser in Milliseconds. Min=275150, Max=275150, Average=275150, Median=275150, Number of runs=1, Individual execution duration=[275150] **
** *** **

Rio Socket for file with 4KB a line:
User with max deletions is /user/mwcl_musicbrainz, count of deletions=237776. Elapsed time=00:04:07.2763483
Requesting to close all call back sockets.
** Printing results of the perf run (C#) **
** Execution time for RunRDDMaxDeletionsByUser in Milliseconds. Min=247276, Max=247276, Average=247276, Median=247276, Number of runs=1, Individual execution duration=[247276] **
** *** **

Traditional Socket for file with 64KB a line:
User with max deletions is /user/mwcl_musicbrainz, count of deletions=5167. Elapsed time=00:01:47.3907642
** Printing results of the perf run (C#) **
** Execution time for RunRDDMaxDeletionsByUser in Milliseconds. Min=107390, Max=107390, Average=107390, Median=107390, Number of runs=1, Individual execution duration=[107390] **
** *** **

Rio Socket for file with 64KB a line:
User with max deletions is /user/mwcl_musicbrainz, count of deletions=5167. Elapsed time=00:01:33.1928019
** Printing results of the perf run (C#) **
** Execution time for RunRDDMaxDeletionsByUser in Milliseconds. Min=93192, Max=93192, Average=93192, Median=93192, Number of runs=1, Individual execution duration=[93192] **
** *** **

Perf results for RIO socket by SocketTest.exe(No spark):
send/recv data without operation: 164.4418838 seconds/10000 messages (10 MB per message)
send/recv data with operation: 1026.1169171 seconds/10000 messages (10 MB per message)

Perf results for tradition socket by SocketTest.exe(No spark):
send/recv data without operation: 1159.0978946 seconds/10000 messages (10 MB per message)
send/recv data with operation: 2555.039326 seconds/10000 messages (10 MB per message)

### DIFF
--- a/csharp/build.sh
+++ b/csharp/build.sh
@@ -1,7 +1,7 @@
 #!/bin/bash
 
 export FWDIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
-
+export CppDll=NoCpp
 export XBUILDOPT=/verbosity:minimal
 
 if [ -z $builduri ];
@@ -37,7 +37,7 @@ export CONFIGURATION=$STEP
 
 export STEP=$CONFIGURATION
 
-xbuild /p:Configuration=$CONFIGURATION $XBUILDOPT $PROJ
+xbuild "/p:Configuration=$CONFIGURATION;AllowUnsafeBlocks=true" $XBUILDOPT $PROJ
 export RC=$? && [ $RC -ne 0 ] && error_exit
 echo "BUILD ok for $CONFIGURATION $PROJ"
 
@@ -46,7 +46,7 @@ export STEP=Release
 
 export CONFIGURATION=$STEP
 
-xbuild /p:Configuration=$CONFIGURATION $XBUILDOPT $PROJ
+xbuild "/p:Configuration=$CONFIGURATION;AllowUnsafeBlocks=true" $XBUILDOPT $PROJ
 export RC=$? && [ $RC -ne 0 ] && error_exit
 echo "BUILD ok for $CONFIGURATION $PROJ"
 
--- a/.gitignore
+++ b/.gitignore
@@ -4,6 +4,7 @@
 *.user
 *.csproj.user
 *.iml
+*.VC.*
 
 # Compiled source #
 ###################
@@ -27,6 +28,10 @@
 **/packages/
 **/target/
 **/.idea/
+**/x64/
+**/debug/
+**/release/
+**/TestResults/
 scala/dependency-reduced-pom.xml
 build/runtime/
 build/tools/
--- a/build/Build.cmd
+++ b/build/Build.cmd
@@ -106,6 +106,29 @@ if EXIST "%CMDHOME%\lib" (
   )
 )
 
+:buildCpp
+Set CppSkipped=0
+@echo Assemble Mobius C++ components (RIOSOCK)
+pushd %CMDHOME%\..\cpp
+call Clean.cmd
+call Build.cmd
+
+if %ERRORLEVEL% EQU 2 (
+  set CppSkipped=1
+  popd
+  goto :buildCSharp
+)
+
+if %ERRORLEVEL% NEQ 0 (
+  @echo Build Mobius C++ RIOSOCK failed, stop building.
+  popd
+  goto :eof
+)
+@echo Mobius C++ RIOSOCK binaries
+copy /y x64\Release\*.dll "%SPARKCLR_HOME%\bin\"
+copy /y x64\Release\*.pdb "%SPARKCLR_HOME%\bin\"
+popd
+
 :buildCSharp
 @echo Assemble Mobius C# components
 pushd "%CMDHOME%\..\csharp"
@@ -215,3 +238,16 @@ copy /Y "%CMDHOME%\..\notes\mobius-release-info.md"
 
 :distdone
 popd
+
+if %CppSkipped% EQU 1 (
+    @echo. 
+    @echo ============================================================================================
+    @echo.
+    @echo   Note!!! Skipped to build Mobius C++ components due to missing VC++ Build Toolset.
+    @echo              If you want to compile C++ components, please enalble VC++ language from
+    @echo              Visual Studio. You can either download "Visual C++ Build Tools" availabe at 
+    @echo              "http://landinghub.visualstudio.com/visual-cpp-build-tools"
+    @echo. 
+    @echo ============================================================================================
+    @echo.
+)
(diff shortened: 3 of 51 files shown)


skaarthik:
@hebinhuang - the numbers are good to see. How real-world-like is the data you used for benchmarking? I am wondering what % of typical workloads will benefit from this improvement. I am aware that it may not be easy to answer that question but that estimate could help us decide whether to keep the traditional socket as the default or not.

hebinhuang:
I have confirmed with Renyi this afternoon. The size of each line is about 4 ~ 5 KB in our internal project. So the file with 4KB a line is close to real-world-like, which has 11% improvement.

csharp/build.sh
```diff
@@ -37,7 +37,7 @@ export CONFIGURATION=$STEP
 
 export STEP=$CONFIGURATION
 
-xbuild /p:Configuration=$CONFIGURATION $XBUILDOPT $PROJ
+xbuild /p:Configuration=$CONFIGURATION;AllowUnsafeBlocks=true $XBUILDOPT $PROJ
```
qualiu: "xbuild /p:Configuration=$CONFIGURATION;AllowUnsafeBlocks=true" can be error on Linux as the semicolon ";" is treated as the separator of two commands.
That's why compilation failed : [./build.sh: line 40: /verbosity:minimal: No such file or directory]( https://travis-ci.org/Microsoft/Mobius/builds/139925856#L2208)

csharp/build.sh
```diff
@@ -46,7 +46,7 @@ export STEP=Release
 
 export CONFIGURATION=$STEP
 
-xbuild /p:Configuration=$CONFIGURATION $XBUILDOPT $PROJ
+xbuild /p:Configuration=$CONFIGURATION;AllowUnsafeBlocks=true $XBUILDOPT $PROJ
```
qualiu: And this line is also need to be corrected with the same reason of line 40.

csharp/build.sh
hebinhuang: @qualiu, does it use "," comma?

qualiu: Probably you can quote them just like : 
xbuild "/p:Configuration=$CONFIGURATION;AllowUnsafeBlocks=true"

Wait a minute, I'll restart my computer to verify that on virtual CentOS.

qualiu: Verified OK on CentOS 7, just add quotes: (sorry take a long time to prepare the env)
xbuild "/p:Configuration=$CONFIGURATION;AllowUnsafeBlocks=true" $XBUILDOPT $PROJ

hebinhuang: @qualiu 👍

hebinhuang: @qualiu, is the quotation mark before /p: or following /p:?

skaarthik:
@hebinhuang - FYI: I cannot build the local clone of your repo in my dev box. It fails because of the missing native build toolset. We need to make the build experience and first-time experience smooth for .NET devs who may not have toolset for building vcxproj.

csharp/build.sh
qualiu: @hebinhuang : just as the latest comment shown, 
quote the intact item as intact : "/p:Configuration=$CONFIGURATION;AllowUnsafeBlocks=true"

hebinhuang:
@skaarthik - does it requires VC++ installed? my dev box has VC++.

skaarthik:
@hebinhuang - yes, it looks like VC++ installation is required to build the code. Can we put checks in place to let the build exclude vcxproj if the toolset is not available in the machine? That also makes a case for making traditional socket the default option (we should make it trivially easy to switch to the optimized implementation though)

Please also review and address the code coverage gaps reported by CodeCov.

qualiu:
@hebinhuang : LooksGoodToMe : Verified(build + run-samples) on Windows/CentOS-7.0/Ubuntu-16.04 by:
```
git clone https://github.com/Microsoft/Mobius.git
git branch hb
git checkout hb
git pull https://github.com/hebinhuang/Mobius.git ManagedRioSock
```

hebinhuang:
Thanks @skaarthik and @qualiu for the review.
I will focus on to add more unit test for the CodeCov. For the build, I will bring a fresh box and see it how it work.

danielli90:
Submitted rebuild at AppVeyor.

skaarthik:
@danielli90 - do we know what caused the previous failure in AppVeyor? Let's not miss the opportunities to identify transient errors caused by the bugs in the code.

danielli90:
AppVeyor keeps track of PR build history. The first failure is an appveyor issue - hence the restart.

skaarthik:
Just checked the previous build error. It says "Exception encountered unloading AppDomain". It does not seem like an AppVeyor issue.

skaarthik:
Looks the other earlier builds failed with the same error message.

danielli90:
yes - "xception encountered unloading AppDomain". @hebinhuang is looking into it.

hebinhuang:
It looks like the error "Exception encountered unloading AppDomain" is caused by the backend threads not terminal.
1. The unit test of SocketWrapperTest has a while(true) thread. I have fixed it.
2. The WeakObjectManagerImp in SparkCLRTestEnvironment is not explicitly disposed. So the thread in WeakObjectManagerImp is always running until GC collect it.

hebinhuang:
Hi @skaarthik,
@danielli90 and I discussed the root cause of the error "Exception encountered unloading AppDomain". After disabled/enabled RIOSocketWrapper unit test, we have found the root cause. The free AppVeyor has only one core. Our Mobius has two backend threads that created by RioNative and WeekObjectManagerImp singleton object. The two backend threads keep running until RioNative and WeekObjectManagerImp object got disposed (process exist).
Nunit will waits for all thread terminated after all test finished. Due to the single core of AppVeyor, the two backend threads cannot be terminated in time, then Nunit hit unloading AppDomain exception.

In the fix, I explicitly calls dispose() of RioNative in the unit test. I have tried two times, and no failure.

danielli90:
LGTM.

skaarthik:
let's not mix maven update with other changes in this PR. I have sent a separate PR with fixes both for AppVeyor and Travis

hebinhuang:
@skaarthik, I will revert it once your PR is merged.

PR #488, opened 2016-07-04
When check and release unused JVMObjectTracker objects, change the speed adaptively according to current objects count.

qintao1976:
When check and release unused JVMObjectTracker objects, change the speed adaptively according to current objects count.


qualiu:
LGTM

PR #490, opened 2016-07-07
Download big file to temp one to avoid possible download failure and use corrupted file.

qualiu:


qualiu:
Download big file may failed, such as running sample as following :
` {Mobius}\build\localmode\RunSamples.cmd`
This will download several files including one big file (like 266M  spark-1.6.2-bin-hadoop2.6.tgz) which will take a long time especially if network's slow.
And if download failed (or killed/canceled), next time run "RunSamples.cmd" again, it'll use the corrupted tgz file (to decompress and use), and consequently cause problem like following (using corrupted jar) :
```
Exception in thread "main" java.lang.NoClassDefFoundError: org/apache/spark/launcher/SparkSubmitArgumentsParser
        at java.lang.ClassLoader.defineClass1(Native Method)
        at java.lang.ClassLoader.defineClass(ClassLoader.java:792)

```
It's a common way to download to a temp position and rename it, as it's easy and cost little.

tawan0109:
LGTM

danielli90:
LGTM. Thanks @qualiu

PR #492, opened 2016-07-08
Rebuild state for DStream UpdateStateByKey operation by replaying batches instead of checkpoint 

tawan0109:
For stateful operation `UpdateStateByKey`, as we know, checkpoint is necessary for RDD lineage truncation and failure recovery by saving state data to reliable storage. However, it’s very challenging to checkpoint large volume data with a reasonable time expense.

In the scenario that the input stream is from Kafka, and the state (data to be checkpointed) can be rebuild by replaying last *N* batches, there is an alternative to Spark checkpoint. This PR tries to provide this alternative solution by combining RDD local checkpoint with offset range checkpoint of `KafkaDynamicRepartitionInputDStream`.

Solution:

* Save Kafka read offset range data into a staging directory`{spark.checkpointDirectory}\kafka-checkpoint\offsets_<batch_time_in_millseconds>.staging` at the beginning of each batch.
 * Kafka read offset range data is saved into a file per topic per DC under the staging directory.
* Commit offset range data by renaming staging directory to `offsets_<batch_time_in_millseconds>` when batch completes successfully. 
* When KafkaDynamicRepartitionInputDStream starts, if `spark.mobius.streaming.localCheckpoint.enabled` is set to true, will first load offset range data for last `spark.mobius.streaming.kafka.replayBatches` batches from `{spark.checkpointDir}\kafka_checkpoint`, merge them for one batch, then repull data from Kafka for replaying.

Options:

* Local checkpoint
 * `spark.mobius.streaming.localCheckpoint.enabled` - whether local checkpoint is enabled, by defaut  `false`.
 * `spark.mobius.streaming.localCheckpoint.replicas` - number of replicas for each block of an local checkpointed RDD. A large number will reduce the possiblity of missing blocks when exectuor lost happens, by default is `3`.
 
* Kafka read offset range checkpoint
 * `spark.mobius.streaming.kafka.enableOffsetCheckpoint` - control whether checkpoint for Kafka read offset is enabled, by default `false`.
 * `spark.mobius.streaming.kafka.maxRetainedOffsetCheckpoints` - how many offset range checkpoints are retained. If this number is less than `spark.mobius.streaming.kafka.replayBatches`, `spark.mobius.streaming.kafka.replayBatches` will be used. By default this number is `0`.

* Kafka read offset ranges replay
 * `spark.mobius.streaming.kafka.replayBatches` - how many batches need to be replayed to rebuild the state.

Build changes:
* As Hadoop related operations are involved in newly added unit tests, need to set `HADOOP_HOME` properly before run `mvn`.

### DIFF
--- a/build/localmode/downloadtools.ps1
+++ b/build/localmode/downloadtools.ps1
@@ -195,6 +195,27 @@ function Untar-File($tarFile, $targetDir)
     Write-Output "[downloadtools.Untar-File] Extraction completed. Time taken: $howlong"
 }
 
+function Download-Winutils
+{
+    $winutilsBin = "$toolsDir\winutils\bin"
+    if (!(test-path "$winutilsBin"))
+    {
+        New-Item -ItemType Directory -Force -Path $winutilsBin | Out-Null
+    }
+
+    $winutilsExe = "$winutilsBin\winutils.exe"
+    if (!(test-path $winutilsExe))
+    {
+        $url = "https://github.com/MobiusForSpark/winutils/blob/master/hadoop-2.6.0/bin/winutils.exe?raw=true"
+        $output=$winutilsExe
+        Download-File $url $output
+    }
+    else
+    {
+        Write-Output "[downloadtools.Download-RuntimeDependencies] $winutilsExe exists already. No download and extraction needed"
+    }
+}
+
 function Download-BuildTools
 {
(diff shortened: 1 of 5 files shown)


xiongrenyi:
LGTM

looks great. thanks a lot for detailed feature description and adding scala unit tests.

PR #498, opened 2016-07-12
Merge bug fix from master to Branch-1.6

qintao1976:
Merge the following bug fix to branch1.6: (1) renyi's fix of CSharpDStream, (2) ReduceByWindowAndKey fix

### DIFF
--- a/build/localmode/downloadtools.ps1
+++ b/build/localmode/downloadtools.ps1
@@ -213,11 +213,11 @@ function Download-BuildTools
     }
     
     # Apache Maven
-	$mvnVer = "apache-maven-3.3.3"
+	$mvnVer = "apache-maven-3.3.9"
     $mvnCmd = "$toolsDir\$mvnVer\bin\mvn.cmd"
     if (!(test-path $mvnCmd))
     {
-        $url = "http://www.us.apache.org/dist/maven/maven-3/3.3.3/binaries/$mvnVer-bin.tar.gz"
+        $url = "http://www.us.apache.org/dist/maven/maven-3/3.3.9/binaries/$mvnVer-bin.tar.gz"
         $output="$toolsDir\$mvnVer-bin.tar.gz"
         Download-File $url $output
         Untar-File $output $toolsDir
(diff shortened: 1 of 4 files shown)


skaarthik:
Can we merge all changes from master to branch 1.6 instead of selected changes?

qintao1976:
There are two reasons:
1) Although originated from the same parent, now Master branch and branch-1.6 are diverse. I have tried to merge all changes from master, but failed due to conflict.
2) In main branch, there are some break change. For example, the time unit of duration for StreamingContext has changed from milliseconds to seconds.

qintao1976:
Maven download failure for the first Appveyor build:

[downloadtools.Download-File] Download completed. Time taken: 515 milliseconds
[downloadtools.Untar-File] Extracting C:\projects\sparkclr\build\localmode\..\tools\apache-maven-3.3.3-bin.tar.gz to C:\projects\sparkclr\build\localmode\..\tools ...
Error extracting C:\projects\sparkclr\build\localmode\..\tools\apache-maven-3.3.3-bin.tar.gz 
Exception message EOS reading GZIP header

skaarthik:
We have to do it eventually anyways at the end of M3. So, I suggest doing it now. We should just take master changes and overwrite any conflicts in 1.6 branch (most likely the conflicts are in version references). We should document the behavior change in streaming API - it is a bug fix after all.

danielli90:
1.6 branch  is snapped to Spark 1.6.1. The master branch is snapped to Spark 1.6.2. Are these break changes " the time unit of duration for StreamingContext has changed from milliseconds to seconds" introduced in Spark 1.6.2? If yes, we may need to keep 1.6 branch aligned with Spark 1.6.1 as-is (and keep the master branch aligned with Spark 1.6.2)

skaarthik:
Master will switch to Spark 2.0 soon. Going with what Spark is doing let's keep 1.6 branch for all Spark 1.6.* releases. In the worst  case we could overload the millisecond API in 1.6 branch for the change made recently and keep the old behavior intact. We can introduce the breaking change in Mobius 2.0.0 release.

qintao1976:
I don't think the change of time unit is introduced by Spark 1.6.2. I think it is only change in C# side. But I am not sure. Perhaps renyi knows the reason. Thanks

xiongrenyi:
Time unit from milliseconds to seconds is actually a bug fix to align with both Scala and python.

xiongrenyi:
LGTM

skaarthik:
Per the discussion offline, let's revert the change to batchwindow in SteamingContext in master branch and bring in all changes from master to branch-1.6. We can then do a release in branch-1.6.

qintao1976:
Close this PR because we need to merge all changes from master to branch-1.6. Need to create a new PR.

PR #507, opened 2016-07-19
unifying KafkaUtils.CreateDirectStream API

xiongrenyi:


csharp/Adapter/Microsoft.Spark.CSharp/Streaming/Kafka.cs
```diff
@@ -170,5 +141,119 @@ public class KafkaUtils
...
+
+        /// <summary>
+        /// topics should contain only one topic if choose to repartitions to a configured numPartitions
+        /// TODO: move to scala and merge into DynamicPartitionKafkaRDD.getPartitions to remove above limitation
+        /// </summary>
+        /// <param name="ssc"></param>
+        /// <param name="topics"></param>
+        /// <param name="kafkaParams"></param>
+        /// <returns></returns>
+        private static int GetNumPartitionsFromConfig(StreamingContext ssc, List<string> topics, Dictionary<string, string> kafkaParams)
```
qintao1976: >GetNumPartitionsFromConfig [](start = 27, length = 26)

Consider to change the name to something like "GetDynamicPartitionNum"?

qintao1976:
LGTM

Please consider to do some validation in cluster before merge the PR.

Thanks

csharp/Adapter/Microsoft.Spark.CSharp/Streaming/Kafka.cs
xiongrenyi: this actually is 'static' version of repartition, maxMessagesPerTask based repartition is 'dynamic'

xiongrenyi:
revert readFunc signature and use GetOffsetRange util instead, which is less impactful and more flexible

tested in cluster

qintao1976:
LGTM

xiongrenyi:
@qintao1976 Can you also verify on your side when you get a chance?

PR #511, opened 2016-07-26
add try finally block to parallel job since driver does not exit if the job fails

xiongrenyi:
@tawan0109 did a comprehensive investigation on hanging SparkSML application after running for a couple of hours and discovered that driver program didn't exit when parallel job threw exception in a different thread which failed to enqueue the rdd for which driver was waiting forever.
add try finally block to always enqueuer, if parallel job failed to materialize the rdd, let main thread retry it as usual.


tawan0109:
LGTM

xiongrenyi:
32% classes vs 30%, 38% lines vs 37% covered 
breakdown sent in email

xiongrenyi:
code coverage is for scala

PR #513, opened 2016-07-29
refactoring to change low level enumerator back to normal

xiongrenyi:
it was changed to low level enumerator API to add some perf metrics which actually can be done in normal enumerator as well.

### DIFF
--- a/csharp/Worker/Microsoft.Spark.CSharp/Worker.cs
+++ b/csharp/Worker/Microsoft.Spark.CSharp/Worker.cs
@@ -254,7 +254,6 @@ namespace Microsoft.Spark.CSharp
             if (lengthOfCommandByteArray > 0)
             {
                 var commandProcessWatch = new Stopwatch();
-                var funcProcessWatch = new Stopwatch();
                 commandProcessWatch.Start();
 
                 int stageId = ReadDiagnosticsInfo(networkStream);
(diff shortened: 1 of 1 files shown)


qualiu:
LGTM  (with following test)

```
test\csharp\testKeyValueStream\test-by-starting-socket.bat -p 9113 -e 10240 -n 200 -d 1 -t 6 2>&1 | lzmw -it "args.\d+|sizeof|final|end of|begin of|exception|OutOfMemory\w*|--\S+-memory" --nt "weak|InterruptedException" -e "elementCount|testTime\w*"

test\csharp\testKeyValueStream\test-by-starting-socket.bat -c chkDir -p 9116 -w 6 -s 1 -e 10240 -n 200 -d 1 -t 6 2>&1 | lzmw -it "args.\d+|sizeof|final|end of|begin of|exception|OutOfMemory\w*|--\S+-memory" --nt "weak|InterruptedException" -e "elementCount|testTime\w*"

test\tools\start-zookeeper-kafka.bat
test\csharp\kafkaStreamTest\test.bat WindowSlideTest -Topics test -d 1 -w 5 -s 1 2>&1 | lzmw -it "args.\d+|sumcount|exception"

test\csharp\kafkaStreamTest\create-2-topics-for-test.bat
test\csharp\kafkaStreamTest\test.bat UnionTopicTest -Topic1 id_count_1 -Topic2 id_count_2 -d 1 -f saveTxtDir -w 5 -s 1 2>&1 | lzmw -it "args.\d+|sumcount|exception"

::## this test need data , such as D:\cosmos\download-stream\tenant\csv-2015-10-01
test\csharp\txtStreamTest\test.bat D:\cosmos\download-stream\tenant\csv-2015-10-01  2>&1 | lzmw -it "exception|RDD count|used time|will read|submit" -e "\d+(\.\d+)?\s*m?s|true|false|rio\w*|x64"

```

qintao1976:
LGTM

csharp/Worker/Microsoft.Spark.CSharp/Worker.cs
```diff
@@ -455,185 +435,91 @@ private static long ToUnixTime(DateTime dt)
...
-                        else
-                        {
-                            result[0] = null;
-                        }
-                        break;
-                    }
 
-                case SerializedMode.Row:
-                    {
-                        Debug.Assert(messageLength > 0);
```
danielli90: I think we should keep Debug.Assert() here, to ensure the logging/assertion characteristics as the same as before, unless we deem this assert() is completely unnecessary.

csharp/Worker/Microsoft.Spark.CSharp/Worker.cs
```diff
@@ -455,185 +435,91 @@ private static long ToUnixTime(DateTime dt)
...
+                    switch (mode)
                     {
-                        result = new object[1];
-                        if (messageLength > 0)
-                        {
-                            byte[] buffer = SerDe.ReadBytes(inputStream, messageLength);
-                            if (buffer == null)
+                        case SerializedMode.String:
                             {
-                                logger.LogDebug("Buffer is null. Message length is {0}", messageLength);
```
danielli90: I think we should keep this logging message, to ensure the logging/assertion characteristics as the same as before, unless we deem this logging line is completely unnecessary.

danielli90:
LGTM, except two minor inline comments suggesting to keep the logging/assertion characteristics as the same as before, unless this specific logging line and assertion are deemed completely unnecessary.

xiongrenyi:
@danielli90  both make sense, added back.

hebinhuang:
LGTM

PR #519, opened 2016-08-05
upgrade to spark 2.0

skaarthik:


skaarthik:
Just testing the changes in this PR. Spark 2.0 upgrade will not happen before M3 release.

PR #521, opened 2016-08-07
Add read and write buffer for CSharpWorker

tawan0109:
For CSharpWorker, in some scenarios adding buffer with proper size can significantly improve the r/w performance using socket, a typical scenario is  that large amount of small records are written to JVM process from CSharpWorker.

In this PR,

* Use a `BufferedStream` to wrap `NetworkStream` to provide buffer functionality. 

 * `BufferedStream` uses one shared buffer for both read and write operation, however, this limits the usage in the scenario that read and write operation are alternated.
 * To avoid this limit, construct two `BufferedStream` instances separately for read and write operation .

* Buffer size for read and write operation can be configured independently by options:

 * `spark.mobius.CSharpWorker.readBufferSize` - Buffer size in bytes for operation of reading data from JVM process, by default is 8KB
 * `spark.mobius.CSharpWorker.writeBufferSize` - Buffer size in bytes for operation of writing data back to JVM process, by default is 8KB
* Example:

> sparkclr-submit.cmd --master yarn-cluster --conf spark.mobius.CSharpWorker.readBufferSize=2097152  --conf spark.mobius.CSharpWorker.writeBufferSize=2097152  ... `


skaarthik:
@tawan0109 - would it be better to use --conf parameter instead of app.config for the newly introduced settings?

tawan0109:
@skaarthik  I considered this approach, but I gave it up for below reasons:
* Options passed by `--conf` can only be accessed via `SparkConf` instance. Buffer size is used in CSharpWorker side, but currently in Mobius `SparkConf` is only available in driver (c#  process).  We can pass SparkConf from Spark execturor(need to change Spark code and carefully define the serialized format) or via `SparkContext.BuildCommand()` (add extra overhead for bytes transferred, and need to materialized SparkConf first)

* Buffer size is considered to be a global variable in CSharpWorker, and should be determined at the start of CSharpWorker execution.

Please let me know whether this makes sense to you.

hebinhuang:
@tawan0109, you can refer to my changes of RioSocket that includes the code to use --conf and adopt the settings at CSharpWorker.  Search "spark.mobius.CSharp.socketType".

https://github.com/Microsoft/Mobius/commit/b6ec24119ea4f38115c9bcf699b9fb614007db60#diff-225710a2890fc0eef1782600427a8956R209

tawan0109:
Thanks @hebinhuang , it's a good idea to use environment variables to pass configuration to CSharpWorker process, I'll make changes to this PR.

csharp/WorkerTest/WorkerTest.cs
```diff
@@ -70,7 +71,7 @@ public void CleanUpSocketWrapper()
         StringBuilder output = new StringBuilder();
         private readonly object syncLock = new object();
 
-        private ISocketWrapper CreateServer(out Process worker)
+        private ISocketWrapper CreateServer(out Process worker, int readBufferSize = 8192, int writeBufferSize = 8192)
```
skaarthik: add notes on why these values are picked as default values

skaarthik:
@tawan0109 - changes look good overall.

Need documentation on the newly introduced Mobius settings. Documentation need to cover when to override the default settings and how to pick a good buffer size.

tawan0109:
@skaarthik  Sure, will add the document.

hebinhuang:
LGTM

csharp/Adapter/Microsoft.Spark.CSharp/Configuration/ConfigurationService.cs
```diff
@@ -21,6 +21,8 @@ internal class ConfigurationService : IConfigurationService
         public const string CSharpWorkerPathSettingKey = "CSharpWorkerPath";
         public const string CSharpBackendPortNumberSettingKey = "CSharpBackendPortNumber";
         public const string CSharpSocketTypeEnvName = "spark.mobius.CSharp.socketType";
+        public const string CSharpWorkerReadBufferSizeEnvName = "spark.mobius.CSharpWorker.readBuffer";
```
skaarthik: Can we use spark.mobius.CSharpWorker.readBufferSize and spark.mobius.CSharpWorker.writeBufferSize?

skaarthik:
LGTM
Only feedback is on the the name of the config settings. Consider renaming for clarity.
Please also update documentation @ https://github.com/Microsoft/Mobius/blob/master/notes/configuration-mobius.md before merge

qualiu:
Test in local mode : found settings not effect :
image

test scripts:
```
d:\msgit\revMobius>
git checkout master 
git pull upstream master
git pull --squash https://github.com/tawan0109/SparkCLR buffered_networkstream


d:\msgit\testMobius>
csharp\update-MobiusCodeRoot-and-project-files.bat d:\msgit\revMobius
csharp\Clean.cmd
csharp\Build.cmd release

set SparkOptions=--num-executors 8 --executor-cores 4 --executor-memory 8G --driver-memory 10G --conf spark.streaming.nao.loadExistingFiles=true --conf spark.streaming.kafka.maxRetries=300 --conf spark.yarn.executor.memoryOverhead=18000 --conf spark.streaming.kafka.maxRetries=20 --jars d:\msgit\revMobius\build\dependencies\spark-streaming-kafka-assembly_2.10-1.6.1.jar --conf spark.mobius.streaming.kafka.CSharpReader.enabled=true --conf spark.mobius.CSharpWorker.readBufferSize=20480 --conf spark.mobius.CSharpWorker.writeBufferSize=20480

csharp\txtStreamTest\test.bat D:\cosmos\download-stream\tenant\csv-2015-10-01 2>&1 | tools\lzmw -P --nt "###" -it "MobiusTest\w+|SparkOption\w*|([\w\.]*\.\w*mobius\w*\.[\w\.]*)|RDD count|used time|exception|(read|write) buffer size:"

```

qualiu:
Again test , but on cluster : also not found effect sign in log 
* I see pull request code it's INFO level but not found in stderr an stdout.
```
 Logger.LogInfo(string.Format("TaskRunner [{0}] is running ..., read buffer size: {1}, write buffer size: {2}", trId, readBufferSize, writeBufferSize));
```

image

image
image

qualiu:
Validate ok now :
```
csharp\txtStreamTest\test.bat D:\cosmos\download-stream\tenant\csv-2015-10-01 2>&1 | tools\lzmw -P --nt "###" -e "\w+\s*=\s*\d+" -it "MobiusTest\w+|SparkOption\w*|([\w\.]*\.\w*mobius\w*\.[\w\.]*)|RDD count|used time|exception|(read|write) buffer size:"
```
image

qualiu:
LGTM

PR #528, opened 2016-08-11
Improve the implementation of RDD TakeOrdered and Top operations

tawan0109:
Current implementation for RDD `TakeOrdered` and `Top` operations is based on in-memory sort, which can't handle the scenario that partition data is too large to fit into memory.

This PR tries to solve this problem by 
* Using a bounded priority queue to find out the top K elements for each partition, priority queue only holds K elements in memory and don't need to sort all data first
* Adding an extra `Reduce` operation to aggregate the result data from each partition.
* Implementing `Top` operation by `TakeOrdered` operation.

### DIFF
--- a/csharp/Adapter/Microsoft.Spark.CSharp/Core/PriorityQueue.cs
+++ b/csharp/Adapter/Microsoft.Spark.CSharp/Core/PriorityQueue.cs
@@ -46,7 +46,7 @@ namespace Microsoft.Spark.CSharp.Core
             var i = elementCount;
             if (i >= queue.Length)
             {
-                if (GE(queue[0], e)) // compare it with root of the heap
+                if (GT(queue[0], e)) // compare it with root of the heap
                 {
                     queue[0] = e;
                     SiftDownHeapRoot();
@@ -77,7 +77,7 @@ namespace Microsoft.Spark.CSharp.Core
                 var child = (k << 1) + 1;
                 var c = queue[child];
                 var right = child + 1;
-                if (right < elementCount && GE(queue[right], c))
+                if (right < elementCount && GT(queue[right], c))
                 {
                     c = queue[child = right];
                 }
@@ -112,10 +112,16 @@ namespace Microsoft.Spark.CSharp.Core
             queue[k] = x;
         }
 
+        // helper method for comparision
+        private bool GT(T a, T b)
+        {
+            return comparer.Compare(a, b) > 0;
+        }
+
         // great or equal, helper method for comparision
         private bool GE(T a, T b)
         {
-            return comparer.Compare(a, b) > 0;
+            return comparer.Compare(a, b) >= 0;
         }
 
         public IEnumerator<T> GetEnumerator()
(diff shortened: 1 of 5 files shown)


csharp/Adapter/Microsoft.Spark.CSharp/Core/PriorityQueue.cs
```diff
@@ -0,0 +1,134 @@
...
+    /// A bounded priority queue implemented with max binary heap.
+    /// 
+    /// Construction steps:
+    ///  1. Build a Max Heap of the first k elements.
+    ///  2. For each element after the kth element, compare it with the root of the max heap,
+    ///    a. If the element is less than the root, replace root with this element, heapify.
+    ///    b. Else ignore it.
+    /// </summary>
+    [Serializable]
+    internal class PriorityQueue<T> : IEnumerable<T> where T : IComparable<T>
```
skaarthik: let's add unit tests for this implementation

csharp/Adapter/Microsoft.Spark.CSharp/Core/RDD.cs
```diff
@@ -1463,33 +1471,65 @@ internal IEnumerable<T> Execute(int pid, IEnumerable<T> input)
...
+            foreach (var e in input)
+            {
+                priorityQueue.Offer(e);
+            }
+
+            return new[] { priorityQueue };
         }
-        internal IEnumerable<T> Execute(int pid, IEnumerable<T> input)
+
+        internal PriorityQueue<T> Execute2(PriorityQueue<T> queue1, PriorityQueue<T> queue2)
```
skaarthik: Consider renaming this method

skaarthik:
LGTM

csharp/Adapter/Microsoft.Spark.CSharp/Core/RDD.cs
tawan0109: @skaarthik will do it in next iteration.

csharp/Adapter/Microsoft.Spark.CSharp/Core/PriorityQueue.cs
tawan0109: @skaarthik this piece of code has already been well covered by current unit tests. Of course I can add some dedicated unit tests for this class in next commit.

skaarthik: @tawan0109 - indirect test coverage is better than not having any coverage at all but not a substitute for unit tests providing direct coverage. Consider this scenario - someone modifying just the implementation of PriorityQueue class need a way to verify their changes and ensure the existing functionality is not broken. Without specific unit tests for this class, there is no straightforward way.  They have to find out the usage of the PriorityQueue, locate the corresponding unit test and rely on them. Even that cannot provide the confidence like direct unit tests. Just wanted to share my 2 cents on this.

csharp/Adapter/Microsoft.Spark.CSharp/Core/RDD.cs
skaarthik: @tawan0109 - feel free to skip it. It is an internal method anyways.

csharp/Adapter/Microsoft.Spark.CSharp/Core/PriorityQueue.cs
tawan0109: @skaarthik  - makes sense to me, added `PriorityQueueTest`.

PR #532, opened 2016-08-12
Escaped special characters in build.cmd

tawan0109:
This PR tries to fix the build error in AppVeyor.


skaarthik:
LGTM

Thanks for fixing this, @tawan0109 . I did not notice the build break in my PR. I incorrectly assumed that the documentation changes in the PR will not cause anything to break :(

PR #537, opened 2016-08-16
Fix error 10022 in Worker.WriteOutput() caused by RioSocket

hebinhuang:
There are random failures on Worker.WriteOutput() with socket error 10022 (An invalid argument was supplied). The reason of the error is because the RioSocketWrapper pass the managed buffer to native buffer and call native code to send the data too many. 

This change is to modify SocketStream.Write with a send data cache, so that to reduce the frequency of calling Riosocket send operation.

### DIFF
--- a/csharp/Adapter/Microsoft.Spark.CSharp/Network/ByteBufChunk.cs
+++ b/csharp/Adapter/Microsoft.Spark.CSharp/Network/ByteBufChunk.cs
@@ -9,6 +9,7 @@ using System.Runtime.CompilerServices;
 using System.Runtime.InteropServices;
 using System.Security;
 using System.Text;
+using Microsoft.Spark.CSharp.Services;
 
 namespace Microsoft.Spark.CSharp.Network
 {
@@ -18,6 +19,7 @@ namespace Microsoft.Spark.CSharp.Network
     /// </summary>
     internal sealed class ByteBufChunk
     {
+        private readonly ILoggerService logger = LoggerServiceFactory.GetLogger(typeof(RioSocketWrapper));
         private readonly Queue<Segment> segmentQueue;
         private readonly int segmentSize;
         private bool disposed;
@@ -286,6 +288,7 @@ namespace Microsoft.Spark.CSharp.Network
                 return;
             }
 
+            logger.LogDebug("Disposing ByteBufChunk [{0}].", ToString());
             if (!IsUnsafe && memory != null)
             {
                 memory = null;
--- a/csharp/Adapter/Microsoft.Spark.CSharp/Network/ByteBufPool.cs
+++ b/csharp/Adapter/Microsoft.Spark.CSharp/Network/ByteBufPool.cs
@@ -2,9 +2,9 @@
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
 using System;
-using System.Collections.Generic;
 using System.Runtime.CompilerServices;
 using System.Text;
+using System.Threading;
 using Microsoft.Spark.CSharp.Services;
 
 namespace Microsoft.Spark.CSharp.Network
@@ -21,6 +21,22 @@ namespace Microsoft.Spark.CSharp.Network
         /// </summary>
         private const int MaxChunkSize = (int)((int.MaxValue + 1L) / 2);
 
+        /// <summary>
+        /// The default segment size to delimit a byte array chunk. 
+        /// </summary>
+        public const int DefaultSegmentSize = 65536; // 64K
+
+        /// <summary>
+        /// The default chunk order used to calculate chunk size and ensure it is a multiple of a segment size.
+        /// </summary>
+        public const int DefaultChunkOrder = 12; // 65536 << 12 = 256 MBytes per chunk
+
+        /// <summary>
+        /// The trial count of trying to allocate buffer from existing ByteBufChunkList.
+        /// The 1000 should be enough to trial.
+        /// </summary>
+        private const int TrialsCount = 1000;
+
         private static readonly Lazy<ByteBufPool> DefaultPool =
             new Lazy<ByteBufPool>(() => new ByteBufPool(DefaultSegmentSize, DefaultChunkOrder, false));
         private static readonly Lazy<ByteBufPool> DefaultUnsafePool =
@@ -29,16 +45,6 @@ namespace Microsoft.Spark.CSharp.Network
         private readonly ILoggerService logger = LoggerServiceFactory.GetLogger(typeof(RioSocketWrapper));
         private readonly bool isUnsafe;
         private readonly ByteBufChunkList qInit, q000, q025, q050, q075, q100;
-
-        /// <summary>
-        /// The default segment size to delimit a byte array chunk. 
-        /// </summary>
-        public const int DefaultSegmentSize = 131072; // 65536; // 64K
-
-        /// <summary>
-        /// The default chunk order used to calculate chunk size and ensure it is a multiple of a segment size.
-        /// </summary>
-        public const int DefaultChunkOrder = 7; // 65536 << 8 = 16 MBytes per chunk
 
         /// <summary>
         /// Initializes a new ByteBufPool instance with a given segment size and chunk order.
@@ -110,13 +116,22 @@ namespace Microsoft.Spark.CSharp.Network
         [MethodImpl(MethodImplOptions.Synchronized)]
         public ByteBuf Allocate()
         {
+            var trial = 0;
             ByteBuf byteBuf;
-            if (q050.Allocate(out byteBuf) || q025.Allocate(out byteBuf) ||
-                q000.Allocate(out byteBuf) || qInit.Allocate(out byteBuf) ||
-                q075.Allocate(out byteBuf))
+            do
             {
-                return byteBuf;
-            }
+                if (q050.Allocate(out byteBuf) || q025.Allocate(out byteBuf) ||
+                    q000.Allocate(out byteBuf) || qInit.Allocate(out byteBuf) ||
+                    q075.Allocate(out byteBuf))
+                {
+                    return byteBuf;
+                }
+
+                // Issues a pause instruction on each loop iteration, 
+                // and not fall back to a more expensive wait on.
+                Thread.Sleep(0); 
+                trial++;
+            } while (trial < TrialsCount);
 
             // Add a new chunk and allocate a segment from it.
             try
@@ -124,9 +139,11 @@ namespace Microsoft.Spark.CSharp.Network
                 var chunk = ByteBufChunk.NewChunk(this, SegmentSize, ChunkSize, isUnsafe);
                 if (!chunk.Allocate(out byteBuf))
                 {
-                    logger.LogError("Failed to allocate a ByteBuf from a new ByteBufChunk - isUnsafe [{0}].", isUnsafe);
+                    logger.LogError("Failed to allocate a ByteBuf from a new ByteBufChunk - isUnsafe [{0}].",
+                        isUnsafe);
                     return null;
                 }
+
                 qInit.Add(chunk);
                 return byteBuf;
             }
--- a/csharp/Adapter/Microsoft.Spark.CSharp/Network/DefaultSocketWrapper.cs
+++ b/csharp/Adapter/Microsoft.Spark.CSharp/Network/DefaultSocketWrapper.cs
@@ -5,6 +5,8 @@ using System;
 using System.IO;
 using System.Net;
 using System.Net.Sockets;
+using Microsoft.Spark.CSharp.Configuration;
+using Microsoft.Spark.CSharp.Services;
 
 namespace Microsoft.Spark.CSharp.Network
 {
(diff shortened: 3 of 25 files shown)


skaarthik:
@hebinhuang - let's validate this fix using the socket streaming as well as FastSML scenarios.

hebinhuang:
This change have fixed the following bugs:
1. Modify SocketStream.Write with a send data cache, so that to reduce the frequency of calling Riosocket send operation and improve performance.
2. Adjust ByteBufChunk size to reduce the frequency of register/unregister RIO buffer, to fix the error 10022 during socket I/O operation (It looks like there is a bug in Windows RIO APIs - the high frequency of Register/unregister RIO buffer caused the error 10022 for socket I/O operation).
3. Separate the background threads to dequeue Send and Receive results. Before this change, we use a single background thread to dequeue Send and Receive results.

This change have been tested on local and yarn cluster mode.

csharp/Adapter/Microsoft.Spark.CSharp/Network/ByteBufPool.cs
```diff
@@ -21,6 +22,21 @@ internal sealed class ByteBufPool
...
+
+        /// <summary>
+        /// The default chunk order used to calculate chunk size and ensure it is a multiple of a segment size.
+        /// </summary>
+        public const int DefaultChunkOrder = 12; // 65536 << 12 = 256 MBytes per chunk
+
+        /// <summary>
+        /// The trial count to try allocation.
+        /// </summary>
+        private const int TrialsCount = 1000;
```
skaarthik: Need more context on this constant as code comments - purpose, reason behind the magic number etc.

csharp/Adapter/Microsoft.Spark.CSharp/Network/ByteBufPool.cs
```diff
@@ -110,23 +116,31 @@ public static ByteBufPool UnsafeDefault
...
             {
-                return byteBuf;
-            }
+                if (q050.Allocate(out byteBuf) || q025.Allocate(out byteBuf) ||
+                    q000.Allocate(out byteBuf) || qInit.Allocate(out byteBuf) ||
+                    q075.Allocate(out byteBuf))
+                {
+                    return byteBuf;
+                }
+                Thread.Sleep(0);
```
skaarthik: why is this needed?

csharp/Adapter/Microsoft.Spark.CSharp/Network/DefaultSocketWrapper.cs
```diff
@@ -80,6 +83,28 @@ public Stream GetStream()
         }
 
         /// <summary>
+        /// Returns a stream used to receive data only.
+        /// </summary>
+        /// <returns>The underlying Stream instance that be used to receive data</returns>
+        public Stream GetInputStream()
+        {
+            var readBufferSize = int.Parse(Environment.GetEnvironmentVariable(ConfigurationService.CSharpWorkerReadBufferSizeEnvName) ?? "8192");
```
skaarthik: Need context on the magic number used here

csharp/Adapter/Microsoft.Spark.CSharp/Network/ISocketWrapper.cs
```diff
@@ -38,6 +38,18 @@ internal interface ISocketWrapper : IDisposable
         Stream GetStream();
```
skaarthik: do we still need this method given that you have added methods for getting input and output stream separately?

csharp/Adapter/Microsoft.Spark.CSharp/Network/RioNative.cs
```diff
@@ -128,25 +129,34 @@ private void Init()
             var maxThreads = 1;
             if (useThreadPool)
             {
-                maxThreads = Environment.ProcessorCount;
+                maxThreads = (Environment.ProcessorCount + 1) >> 1;
```
skaarthik: why?

csharp/Adapter/Microsoft.Spark.CSharp/Network/SocketStream.cs
```diff
@@ -87,10 +88,29 @@ public override long Position
         }
 
         /// <summary>
-        /// Flushes data from the stream.  This is meaningless for us, so it does nothing.
+        /// Flushes data in send cache to the stream.
         /// </summary>
         public override void Flush()
         {
+            if (sendDataCache != null && sendDataCache.IsReadable())
```
skaarthik: do we need to handle the case when IsReadable is false?

csharp/Adapter/Microsoft.Spark.CSharp/Proxy/Ipc/StreamingContextIpcProxy.cs
```diff
@@ -306,7 +305,8 @@ private void ProcessCallbackRequest(object socket)
                                 logger.LogDebug("receive close cmd from Scala side");
                                 break;
                             }
-                            else if (cmd == "callback")
```
skaarthik: what is the reason for this change?

csharp/Worker/Microsoft.Spark.CSharp/Worker.cs
```diff
@@ -343,12 +343,20 @@ private static IFormatter ProcessCommand(Stream inputStream, Stream outputStream
...
+                    {
+                        logger.LogError("WriteOutput() failed at iteration {0}", count);
+                        throw;
+                    }
                     count++;
                     funcProcessWatch.Start();
                 }
 
-                logger.LogDebug("Output entries count: " + count);
+                logger.LogInfo("Output entries count: " + count);
```
skaarthik: what is the reason for this change?

skaarthik:
@hebinhuang - code cov number is dropping in this PR. Please address that.

csharp/Adapter/Microsoft.Spark.CSharp/Network/ByteBufPool.cs
hebinhuang: Thread.Sleep(0) issues a pause instruction on each loop iteration, and not fall back to a more expensive wait on.

csharp/Adapter/Microsoft.Spark.CSharp/Network/ByteBufPool.cs
hebinhuang: Added

csharp/Adapter/Microsoft.Spark.CSharp/Network/SocketStream.cs
hebinhuang: No. If IsReadable is false, it means no data need to be send out.

csharp/Adapter/Microsoft.Spark.CSharp/Proxy/Ipc/StreamingContextIpcProxy.cs
hebinhuang: else if (cmd == "callback") is redundant code. I got the suggestion from ReSharp tool.

csharp/Worker/Microsoft.Spark.CSharp/Worker.cs
hebinhuang: Prompt the log level here, which give us information about the output entries. By default, LogDebug is not shown in the log.

csharp/Adapter/Microsoft.Spark.CSharp/Network/RioNative.cs
hebinhuang: We separate the work threads to process send and receive completion at the below code. So just make the total number of threads still equals to the machine's processor count.

csharp/Adapter/Microsoft.Spark.CSharp/Network/ISocketWrapper.cs
hebinhuang: Tao Wang introduced the input and output stream at the CSharpWorker only. I do not do lots of changes on drive side. So still keep the function here.

csharp/Adapter/Microsoft.Spark.CSharp/Network/DefaultSocketWrapper.cs
hebinhuang: This one I moved from Tao wang's changes. But I have added the comments.

hebinhuang:
@skaarthik , the total number is dropped by 0.34%. The reason is that this PR introduced a new Completion Queue for send completion in native code, and added error handler for the new Completion Queue. So the decreased number is from the error handler in native code.

skaarthik:
@hebinhuang - lines 235 - 240 in SocketStream.cs are newly added code and not covered by any tests. BTW, with pure unit tests, even error handling code can be covered easily.

csharp/Adapter/Microsoft.Spark.CSharp/Network/ByteBufPool.cs
skaarthik: I guess the question is why do we need a pause instruction? Can you add your response as code comments also?

hebinhuang:
The lines you mentioned are already covered by TestRioSocket in SocketWrapperTest.

See the coverage https://codecov.io/github/Microsoft/Mobius/commit/0fe576daad96b2318aabf41127a67c88326e4f3d, the code have been covered.

csharp/Adapter/Microsoft.Spark.CSharp/Network/RioNative.cs
skaarthik: Do we know how this work in YARN?

csharp/Worker/Microsoft.Spark.CSharp/Worker.cs
skaarthik: Right, but how is this "info" useful other than for debugging?

skaarthik:
@hebinhuang - those lines are not covered. Look at the snapshot from CodeCov below

image

csharp/Adapter/Microsoft.Spark.CSharp/Network/ByteBufPool.cs
hebinhuang: Added.

csharp/Adapter/Microsoft.Spark.CSharp/Network/RioNative.cs
hebinhuang: These code is only for Multithread Worker. In simple worker, there are only two threads (separately for send and receive completion). No difference between local and yarn cluster mode.

skaarthik: How will this calculation for maxThreads work within a YARN container where it is possible to control the number of cores?

csharp/Worker/Microsoft.Spark.CSharp/Worker.cs
hebinhuang: this "info" give us a sense of total entries be processed by a worker, just like the "info" for "func process time". Based on the "info" of "Output entries count" and "func process time", we can have the idea of processing time of each process.
So I still think it is better to use "info" than "Debug".

hebinhuang:
Now, it is covered by last merge.

image

csharp/Adapter/Microsoft.Spark.CSharp/Network/RioNative.cs
hebinhuang: Just print out the processor number from Yarn cluster:
CSharpWorker [11500] is running on a machine with 32 processors

In Multithread worker mode, if YARN arrange multiple executors into one machine, it will start multiple CsharpWorker - so the total threads will be "processors X csharpworkers".

csharp/Adapter/Microsoft.Spark.CSharp/Network/ByteBufChunk.cs
```diff
@@ -286,6 +288,7 @@ private void Dispose(bool disposing)
                 return;
             }
 
+            logger.LogDebug("Disposing ByteBufChunk [{0}].", RuntimeHelpers.GetHashCode(this).ToString("X"));
```
qualiu: The HashCode will not change not only in all the threads of the process, but also the processes every time started.
So, if you want to add a log line and print extra info, maybe it's more useful to print the object other than the invariant HashCode.

qualiu:
LGTM

csharp/Adapter/Microsoft.Spark.CSharp/Network/ByteBufChunk.cs
hebinhuang: Changed to the below:

logger.LogDebug("Disposing ByteBufChunk [{0}].", ToString());

csharp/Adapter/Microsoft.Spark.CSharp/Network/RioNative.cs
hebinhuang: As your mentioned, we can get the executor cores in cluster/standalone mode.

CSharpWorker [6260] is started with cores: 28

hebinhuang:
@skaarthik , I have update the PR.

skaarthik:
LGTM

PR #538, opened 2016-08-16
updating examples to use latest NuGet package

skaarthik:


qualiu:
Found 2 lines in  `skaarthikMobius\notes\mobius-release-info.md` need to update .
```
git clone -b m3exampleupdate https://github.com/skaarthik/Mobius skaarthikMobius
lzmw -rp skaarthikMobius -x "1.6.100"
```
search / preview-replace/replace as following:
image

skaarthik:
@qualiu what you are pointing is not a real reference to latest Mobius version. It is used to explain versioning policy. So it is okay not to use latest version there

qualiu:
@skaarthik : ok.  
LGTM

PR #556, opened 2016-08-30
Use logInfo and logError method instead of println

Myasuka:
Finish TODO "logError does not work now" and "TODO: change println to log"

### DIFF
--- a/scala/src/main/org/apache/spark/api/csharp/CSharpBackend.scala
+++ b/scala/src/main/org/apache/spark/api/csharp/CSharpBackend.scala
@@ -7,15 +7,16 @@ package org.apache.spark.api.csharp
 
 import java.io.{DataOutputStream, File, FileOutputStream, IOException}
 import java.net.{InetAddress, InetSocketAddress, ServerSocket, Socket}
-import java.util.concurrent.{LinkedBlockingQueue, BlockingQueue, TimeUnit}
+import java.util.concurrent.{BlockingQueue, LinkedBlockingQueue, TimeUnit}
 
 import io.netty.bootstrap.ServerBootstrap
 import io.netty.channel.nio.NioEventLoopGroup
 import io.netty.channel.socket.SocketChannel
 import io.netty.channel.socket.nio.NioServerSocketChannel
-import io.netty.channel.{ChannelInitializer, EventLoopGroup, ChannelFuture}
+import io.netty.channel.{ChannelFuture, ChannelInitializer, EventLoopGroup}
 import io.netty.handler.codec.LengthFieldBasedFrameDecoder
 import io.netty.handler.codec.bytes.{ByteArrayDecoder, ByteArrayEncoder}
+import org.apache.spark.internal.Logging
 
 
 /**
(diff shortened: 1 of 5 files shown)


scala/src/main/org/apache/spark/api/csharp/CSharpBackendHandler.scala
```diff
@@ -176,31 +174,27 @@ class CSharpBackendHandler(server: CSharpBackend) extends SimpleChannelInboundHa
...
         val jvmObjName = jvmObj match {
           case Some(jObj) => jObj.getClass.getName
           case None => "NullObject"
         }
-        // scalastyle:off println
-        println(s"[CSharpBackendHandler] $methodName on object of type $jvmObjName failed")
-        println(e.getMessage)
-        println(e.printStackTrace())
+        logError(s"On object of type $jvmObjName failed ${e.getMessage}")
+        e.printStackTrace()
```
qualiu: Why not put the error stack into logger ? this may separate it with above log.

Myasuka: I follow the logic of original code and even in Spar source code `printStackTrace()` method is not contained in logger, such as: [Client.scala](https://github.com/apache/spark/blob/v2.0.0/core/src/main/scala/org/apache/spark/deploy/Client.scala#L142)

scala/src/main/org/apache/spark/deploy/csharp/CSharpRunner.scala
```diff
@@ -123,22 +121,23 @@ object CSharpRunner {
           closeBackend(csharpBackend)
         } catch {
           case t: Throwable =>
-            println("[CSharpRunner.main]" + t.getMessage + "\n" + t.getStackTrace)
+            logError(s"${t.getMessage} \n ${t.getStackTrace}")
```
qualiu: I saw there're different stack logging , such as here (stack into logger) to above mentioned (stack may outside)

scala/src/main/org/apache/spark/api/csharp/CSharpBackendHandler.scala
qualiu: As following stack into logger (several places), why not keep only one style since we have our own different stack logging from Apache-Spark (Spark doesn't seem to be a well standard for this point)

qualiu:
LGTM

Myasuka:
@qualiu I have removed redundant printStackTrace method.

scala/src/main/org/apache/spark/api/csharp/CSharpBackend.scala
```diff
@@ -7,15 +7,16 @@ package org.apache.spark.api.csharp
...
 
 import io.netty.bootstrap.ServerBootstrap
 import io.netty.channel.nio.NioEventLoopGroup
 import io.netty.channel.socket.SocketChannel
 import io.netty.channel.socket.nio.NioServerSocketChannel
-import io.netty.channel.{ChannelInitializer, EventLoopGroup, ChannelFuture}
+import io.netty.channel.{ChannelFuture, ChannelInitializer, EventLoopGroup}
 import io.netty.handler.codec.LengthFieldBasedFrameDecoder
 import io.netty.handler.codec.bytes.{ByteArrayDecoder, ByteArrayEncoder}
+import org.apache.spark.internal.Logging
```
skaarthik: We are reusing Logging class in Spark. Is there a way to configure Mobius logging in Scala code independent of Spark's log configuration?

Myasuka: I'm not sure whether having understood what your mean, if you refer to configure the log level such as from `INFO` to `DEBUG`, you can append `--files log4j.properties` in submit arguments on YARN. Or if you refer to configure different from `$SPARK_CONF/log4j.properties`, we can create a new class, for example `MobiusLogging`, extends `Logging` and override the `initializeLogIfNecessary(isInterpreter: Boolean)` method.

skaarthik: I was wondering if there is a way to independently configure Mobius log. I guess we should be able to do filtering to redirect just Mobius logs.

LGTM

PR #557, opened 2016-08-30
[Mobius-102] Add missing license text in batch and powershell scripts

Myasuka:


qualiu:
How did you do it for these 29 files ? manual or script/tool ?

Myasuka:
@qualiu just manually since it does not take much time and the headers of different files are not the same.

But I have to say it's a bit lumbering to manually modify these files.

qualiu:
@Myasuka :+1:  That's quite a careful boring checking/adding. 
I'd considered to use [one tool](https://github.com/qualiu/lzmw) to do it once for all , and for future if there's license lost. 

So that everyone can check/add it very fast (in several seconds) and keep consistency. (Because of the same command lines, maybe 3~9 lines for all kinds of files and license text)

qualiu:
@Myasuka : Have you checked the something extra such as line endings ? especially for the Linux shell files?

qualiu:
@Myasuka :Have you checked something extra such as line endings ? especially for the Linux shell files?

Myasuka:
@qualiu the `.gitattributes` file has configured that text files are LF normalization, so my commits would not change the original logic.

qualiu:
LGTM

skaarthik:
LGTM

PR #567, opened 2016-09-28
Use copy task instead of OS commands so it builds on OSX

shakdoesgithub:
cp -uv is run on OSX, but the option -u is not available in OSX, so I'm using the copy task which works cross operating system. It lets msbuild take care of operation system copying functionality.


skaarthik:
@shakdoesgithub thanks for your contribution. 
"DependsOn" does not seem like a valid attribute to Target element. https://msdn.microsoft.com/en-us/library/t50z2hka.aspx and that is failing the build of your PR in AppVeyor/Windows. It seems to work find in Travis/Linux builds though.

shakdoesgithub:
Thanks for the help @skaarthik, doing another build now with the updated attribute.

PR #573, opened 2016-10-04
use forward slash to work in linux/osx

shakdoesgithub:
Sparks Jar ClassPath doesn't work in in linux because of back slash.  This fixes that.


skaarthik:
Thanks for your contribution @shakdoesgithub

PR #579, opened 2016-10-11
fixing the issue in YARN clusters introduced during Spark 2.0 upgrade

skaarthik:
Adding "yarn" as a valid value for master config - this is allowed in Spark 2.0
Fixing the issues in launching CSharpWorker.exe in YARN clusters - introduced during Spark 2.0 upgrade


tawan0109:
LGTM, thanks for fixing this.

skaarthik:
Thanks @tawan0109

PR #602, opened 2016-12-02
Fix ReduceByKey failure when the value type is a reference type

medvekoma:
ReduceByKey fails with a NPE if the value type is a reference type, even if there are no null values in the PairRDD. 

The reason of the error is that ReduceByKey calls CombineByKey with a seed generator of () => default(V). When the value is a reference type, this seed value is null. The lambda will try to accumulate the values to this seed value of null. A unit test was added to demonstrate this behavior.

The fix not only avoids NPE errors, but it is also more efficient, because it calls the lambda of the ReduceByKey for actual value pairs only.

This behavior is also in line with the Scala and Java execution: ReduceByKey doesn't call the lambda with the default value of the data type.

### DIFF
--- a/csharp/Adapter/Microsoft.Spark.CSharp/Core/PairRDDFunctions.cs
+++ b/csharp/Adapter/Microsoft.Spark.CSharp/Core/PairRDDFunctions.cs
@@ -97,7 +97,11 @@ namespace Microsoft.Spark.CSharp.Core
         /// <returns></returns>
         public static RDD<KeyValuePair<K, V>> ReduceByKey<K, V>(this RDD<KeyValuePair<K, V>> self, Func<V, V, V> reduceFunc, int numPartitions = 0)
         {
-            return CombineByKey(self, () => default(V), reduceFunc, reduceFunc, numPartitions);
+            var locallyCombined = self.MapPartitionsWithIndex(new GroupByMergeHelper<K, V>(reduceFunc).Execute, true);
+
+            var shuffled = locallyCombined.PartitionBy(numPartitions);
+
+            return shuffled.MapPartitionsWithIndex(new GroupByMergeHelper<K, V>(reduceFunc).Execute, true);
         }
 
         /// <summary>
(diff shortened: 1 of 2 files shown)


medvekoma:
I think that the travis error is caused by the version of the nunit assembly on the build server. Can you verify it please?

skaarthik:
Here is the error in Travis log:
```
System.ArgumentException: Argument must be between 0 and 2.
Parameter name: fieldCount
  at System.Version.ToString (System.Int32 fieldCount) [0x0008b] in <8f2c484307284b51944a1a13a14c0266>:0 
  at NUnit.Engine.Services.TestAgency.LaunchAgentProcess (NUnit.Engine.TestPackage package) [0x001d9] in <a2c7edfe381f45afbaecb1dd27d35998>:0 
  at NUnit.Engine.Services.TestAgency.CreateRemoteAgent (NUnit.Engine.TestPackage package, System.Int32 waitTime) [0x00000] in <a2c7edfe381f45afbaecb1dd27d35998>:0 
  at NUnit.Engine.Services.TestAgency.GetAgent (NUnit.Engine.TestPackage package, System.Int32 waitTime) [0x00000] in <a2c7edfe381f45afbaecb1dd27d35998>:0 
  at (wrapper remoting-invoke-with-check) NUnit.Engine.Services.TestAgency:GetAgent (NUnit.Engine.TestPackage,int)
  at NUnit.Engine.Runners.ProcessRunner.LoadPackage () [0x0003a] in <a2c7edfe381f45afbaecb1dd27d35998>:0 
==== Test failed ====
```

skaarthik:
Looks like build also failed in Travis. I see the following message:

/bin/sh: 1: java: not found
[ERROR] Failed to execute goal org.scalatest:scalatest-maven-plugin:1.0:test (test) on project spark-clr_2.11: There are test failures -> [Help 1]

medvekoma:
Thanks @skaarthik for the quick review.

PR #648, opened 2017-03-28
WordCount Example (Batch) exe path correction

kmleow:
It was pointing to SparkClrPi.exe instead of SparkClrWordCount.exe


skaarthik:
Thanks @kmleow 
We will get this merged once a build issue is fixed

PR #694, opened 2018-09-11
Fixed integer Data type issue for UDF

sutyag:
int is not valid type in spark, so fixed it to integer

### DIFF
--- a/csharp/Adapter/Microsoft.Spark.CSharp/Sql/Functions.cs
+++ b/csharp/Adapter/Microsoft.Spark.CSharp/Sql/Functions.cs
@@ -919,7 +919,7 @@ namespace Microsoft.Spark.CSharp.Sql
             {typeof(double), "double"},
             {typeof(float), "float"},
             {typeof(byte), "tinyint"},
-            {typeof(int), "int"},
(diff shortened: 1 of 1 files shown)


skaarthik:
@sutyag thanks for the PR. Can you include a pointer to Spark documentation in your PR that confirms "int is not valid type in spark"? It will be helpful to those who are interested in this PR and make the review faster.

sutyag:
@skaarthik I tried with few samples for UDFs. When i try to register below function;

```
int UDF1(string a){
return 10;
}

it is failing with below error:
Name of the function to register UDF1, method info
[2018-09-17 16:54:58,218] [1] [ERROR] [Microsoft.Spark.CSharp.Interop.Ipc.JvmBridge] - JVM method execution failed: Static method fromJson failed for class org.apache.spark.sql.types.DataType when called with 1 parameters ([Index=1, Type=String, Value="int"], )
[2018-09-17 16:54:58,218] [1] [ERROR] [Microsoft.Spark.CSharp.Interop.Ipc.JvmBridge] - java.util.NoSuchElementException: key not found: int

while method 
string UDF2(int a){
return "10";
} is working without any issue.

```

PR #695, opened 2018-10-03
README: Twitter badge fix

davidk:
This change urlencodes the ?text= parameter so that the button can be used as intended.

I've also noticed that the issue stats buttons for pull request / issues are broken. 

I've left this untouched, so that other Hacktoberfest participants may PR against it, if permissible (for removal/replacement/triage).


skaarthik:
Thanks @davidk for your first-contribution to Mobius.

PR #697, opened 2018-11-13
Dev/sutyag/upgrade mobius

sutyag:
1)	Mobius changes for spark 2.3.1
2)	Fixed Bug: int return type for UDF
3)	Fixed Bug; Duplicate arguments value in UDF
4)	Added Support for Multiple UDFs
5)	Added sample test cases
6)    Fixed Unit test cases


skaarthik:
@sutyag AppVeyor build gets stuck with the following message:

> [SparkCLRSamples.exe.PrintLogLocation] Logs by SparkCLR and Apache Spark are available at C:\Users\appveyor\AppData\Local\Temp\1\SparkCLRLogs
> 9906
> [SparkCLRSamples.exe.Main] Completed running samples. Calling SparkContext.Stop() to tear down ...
> 9907
> [SparkCLRSamples.exe.Main] If this program (SparkCLRSamples.exe) does not terminate in 10 seconds, please manually terminate java process launched by this program!!!

Sparkcontext is not terminating Java process. Do you see this issue when running examples manually?

csharp/Adapter/Microsoft.Spark.CSharp/Interop/Ipc/IJvmBridge.cs
```diff
@@ -12,7 +12,7 @@ namespace Microsoft.Spark.CSharp.Interop.Ipc
     /// <summary>
     /// Behavior of the bridge used for the IPC interop between JVM and CLR
     /// </summary>
-    internal interface IJvmBridge : IDisposable
+    public interface IJvmBridge : IDisposable
```
skaarthik: what is the reason to make this public?

